PR #1, opened 2020-08-10
[WIP] Add version checking

auxesis:


auxesis:
Codebase has changed significantly enough this PR doesn't make sense. 

Will create another when ready.

PR #2, opened 2020-08-18
Add deploy command

pavelnikolov:
Add deploy command

### DIFF
--- a/commands/deploy.go
+++ b/commands/deploy.go
@@ -1,20 +1,144 @@
 package commands
 
 import (
+	"archive/tar"
+	"compress/gzip"
 	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
 	"os"
+	"path/filepath"
+	"strings"
+	"time"
 )
 
-// DeployCmd handles deploying an app to Section
-type DeployCmd struct{}
+// MaxFileSize is the tarball file size allowed to be uploaded in bytes.
+const MaxFileSize = 1073741824 // 1GB
+
+// DeployCmd handles deploying an app to Section.
+type DeployCmd struct {
+	Debug     bool
+	ServerURL string `default:"https://api.section.io/code-upload/v1"`
+}
 
 // Run deploys an app to Section's edge
 func (a *DeployCmd) Run() (err error) {
-	path, err := os.Getwd()
-	if err != nil {
-		panic(err)
+	if a.Debug {
+		fmt.Println("Server URL:", a.ServerURL)
 	}
+	path := "./"
+	var files []string
+	err = filepath.Walk(path, func(path string, info os.FileInfo, err error) error {
+		if info.IsDir() {
+			return nil
+		}
+		if strings.Contains(path, ".git/") {
+			return nil
+		}
+		if strings.Contains(path, ".lint/") {
+			return nil
+		}
+		files = append(files, path)
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("failed to walk path: %v", err)
+	}
+	if a.Debug {
+		for _, file := range files {
+			fmt.Println(file)
+		}
+	}
+
 	fmt.Printf("Packaging %s\n", path)
+
+	tempFile, err := ioutil.TempFile("", "section")
+	if err != nil {
+		return fmt.Errorf("couldn't create a temp file: %v", err)
+	}
+	defer os.Remove(tempFile.Name())
+	err = CreateTarball(tempFile, files)
+	if err != nil {
+		return fmt.Errorf("failed to pack files: %v", err)
+	}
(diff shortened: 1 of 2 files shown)


commands/deploy.go
```diff
@@ -1,20 +1,74 @@
...
+	path := "./"
+	var files []string
+	err = filepath.Walk(path, func(path string, info os.FileInfo, err error) error {
+		if info.IsDir() {
+			return nil
+		}
+		if strings.Contains(path, ".git/") {
+			return nil
+		}
+		if strings.Contains(path, ".lint/") {
```
jstangroome, line 39: Might be more general purpose to have an optional exclude pattern which defaults to anything beginning with a dot `.` by default.

commands/deploy.go
```diff
@@ -1,20 +1,74 @@
...
 	if err != nil {
-		panic(err)
+		return fmt.Errorf("failed to walk path: %v", err)
+	}
+	for _, file := range files {
+		fmt.Println(file)
 	}
+
 	fmt.Printf("Packaging %s\n", path)
+	var buf bytes.Buffer
```
jstangroome: A temp file would be better otherwise this tool will consume as much ram as the size of their gzipped build output

jstangroome (approved):

commands/deploy.go
pavelnikolov: I can also use an `io.Pipe` and directly gzip to the wire and send the output to the server. Then it will not consume much ram and will not consume much disk space (if they are connected). I can change it to a temp file.

commands/deploy.go
pavelnikolov: I am planning to add `.section.json` config file or maybe a `.section/config.json` and `\.*+` would match that perfectly.

commands/deploy.go
jstangroome: Pipe is good too. Wasn't sure if you wanted to validate final tar before beginning upload (e.g. no file read errors, within size limit).

commands/deploy.go
jstangroome: What would such a json file contain?

pavelnikolov: The environment ID. What if the user has multiple projects for multiple envs? How do we know which one we are deploying?

commands/deploy.go
pavelnikolov: Added max file size validation

commands/deploy.go
pavelnikolov: @manibatra just told me that I only needed to use account ID and not env ID. So I will not add a config file per project for now.

commands/deploy.go
```diff
@@ -2,19 +2,91 @@ package commands
...
+type DeployCmd struct {
+	Debug     bool
+	ServerURL string `default:"https://api.section.io/code-upload/v1"`
+}
 
 // Run deploys an app to Section's edge
 func (a *DeployCmd) Run() (err error) {
-	path, err := os.Getwd()
+	fmt.Println("Server URL", a.ServerURL)
+	// path, err := os.Getwd()
```
auxesis: Let's just remove these lines. 

If we need them again, revert the commit, and submit a new PR.

commands/deploy.go
```diff
@@ -2,19 +2,91 @@ package commands
...
+const MaxFileSize = 1073741824 // 1GB
+
+// DeployCmd handles deploying an app to Section.
+type DeployCmd struct {
+	Debug     bool
+	ServerURL string `default:"https://api.section.io/code-upload/v1"`
+}
 
 // Run deploys an app to Section's edge
 func (a *DeployCmd) Run() (err error) {
```
auxesis, line 26: Just noticed that I've used `a` as the variable name for the receiver throughout all the commands. 

This should probably be named `c` (for command). 

Don't bother changing that here, and I'll submit a PR that renames the receiver everywhere.

commands/tar.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"archive/tar"
+	"compress/gzip"
+	"fmt"
+	"io"
+	"os"
+)
+
+// CreateTarball creates a tarball containing all the files in filePaths and writes it to w.
+func CreateTarball(w io.Writer, filePaths []string) error {
```
auxesis: I'm a little reluctant to have these functions in a separate file, because: 

- It seems like it would be better placed in `deploy.go`
- I would expect to see a command for every file in `commands/`, and there isn't a command here

What's the rationale?

commands/deploy.go
pavelnikolov: I will leave this for now

commands/deploy.go
pavelnikolov: done

PR #5, opened 2020-08-22
Add basic analytics with Heap, to report what commands are being run

auxesis:
Gives us an idea about how people are running the CLI in the wild. 

Will need to add support for opting out of analytics.


analytics/analytics.go
```diff
@@ -0,0 +1,102 @@
...
+	Name       string
+	Properties map[string]string
+}
+
+// Identity tries to determine the identity of:
+//
+// - the person running the cli
+// - the machine the cli is being run on
+func identity() (id string) {
+	id, err := gitconfig.Email()
```
pavelnikolov: This should be prohibited without the explicit consent of the user. It is PII. It is completely illegal in Europe.

pavelnikolov: Not to mention that some users might oppose any attempt to being tracked

pavelnikolov:
Most (if not all) of these stats we are going to get from our metrics from the upload API. I personally do not see the need for this 3rd party analytics tool.

analytics/analytics_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/stretchr/testify/assert"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+func TestAnalyticsSubmitPostsToHeap(t *testing.T) {
+	assert := assert.New(t)
+	called := false
```
pavelnikolov: (nitpick) Use the zero value, as Rob Pike says in his proverbs:
`var called bool` is the same.

(Just my personal preference, nothing wrong with your code).

analytics/analytics.go
auxesis: 💯

There's a missing part to this PR – an explicit opt-in. 

I'm thinking I might just simplify this for now and only use the anonymized machine id.

analytics/analytics_test.go
auxesis: I prefer your suggested syntax too. Updating now.

auxesis:
> Most (if not all) of these stats we are going to get from our metrics from the upload API. 

Not quite. 

We can derive some of this from the API calls, but we won't see: 

- How the CLI being used
- CLI usability problems 

Ideally we'll only need these extra insights in the early stages of developing the CLI, while we're working out how to make a great experience. Removing the analytics functionality altogether is definitely something we will consider in the future. 

To reiterate – _any analytics should be explicitly opt-in._ I need to add that explicit opt-in before this PR can be considered mergeable.

analytics/analytics.go
```diff
@@ -0,0 +1,102 @@
...
+		Identity:   identity(),
+		Event:      e.Name,
+		Timestamp:  time.Now(),
+		Properties: e.Properties,
+	}
+	ev := heapTrack{
+		AppID:  HeapAppID,
+		Events: []heapTrackEvent{hte},
+	}
+	json, _ := json.Marshal(ev)
```
pavelnikolov: You can have an error if you encounter memory issues. Don't ignore the error.

analytics/analytics.go
```diff
@@ -0,0 +1,103 @@
...
+
+	uri := fmt.Sprintf("%s/api/track", HeapBaseURI)
+	resp, err := http.Post(uri, "application/json", bytes.NewBuffer(json))
```
pavelnikolov: Please, use `http.Client` and call its `Do` method. The above uses the default client which doesn't have a timeout. If there is a problem it might hang forever. You need a relatively short timeout.

analytics/analytics.go
auxesis: Fixed in 7d27171.

analytics/analytics.go
auxesis: Fixed in 639998478ad8755be36067b206e33fbf709ea49e.

auxesis:
@pavelnikolov have reworked this PR significantly – would love for you to take another look when you have a moment.

Makefile
```diff
@@ -19,7 +19,7 @@ gostaticcheck:
 	staticcheck ./...
 
 goerrcheck:
-	errcheck -exclude .lint/errcheck-excludes -ignoretests ./...
+	errcheck -exclude .lint/errcheck-excludes -blank -ignoretests -ignore 'github.com/section/section-cli/analytics:^Print' ./...
 
 build:
-	go build -o bin/section section.go
+	go build -ldflags "-X 'github.com/section/section-cli/analytics.HeapAppID=892134159'" -o bin/section section.go
```
pavelnikolov, line 25: Can it be just `HeapAppID=892134159`?

analytics/analytics.go
```diff
@@ -0,0 +1,258 @@
...
+	"github.com/alecthomas/kong"
+	"github.com/denisbrodbeck/machineid"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"os/user"
+	"path/filepath"
+	"strings"
+	"time"
```
pavelnikolov, line 16: (nitpick) please, separate the stdlib dependencies and the third party ones with an empty line.

analytics/analytics.go
```diff
@@ -0,0 +1,258 @@
...
+)
+
+var (
+	// HeapBaseURI is the location of Heap API endpoint
+	HeapBaseURI = "https://heapanalytics.com"
+	// HeapAppID identifies what Heap App events are recorded against
+	HeapAppID = "4248790180"
```
pavelnikolov: Doesn't this come from the build?

analytics/analytics.go
```diff
@@ -0,0 +1,258 @@
...
+	// consentPath is the path on disk to where analytics consent is recorded
+	consentPath string
+	// ConsentGiven records whether tracking consent has been given by the user
+	ConsentGiven bool
+	// out is the buffer to write feedback to users
+	out io.Writer
+	// in is the buffer to read feedback from users
+	in io.Reader
```
pavelnikolov, line 34: You can easily avoid global variables and pass everything as params.

analytics/analytics.go
```diff
@@ -0,0 +1,258 @@
...
+	Printf("We'd like to collect some anonymous info about how you use the CLI.\n\n")
+	Printf("Are you OK with this? [y/N] ")
+
+	reader := bufio.NewReader(in)
+	text, err := reader.ReadString('\n')
+	if err != nil {
+		Println("Error: unable to read your response. Exiting.")
+		os.Exit(2)
+	}
+	text = strings.Replace(text, "\n", "", -1) // convert CRLF to LF
```
pavelnikolov: Shouldn't you replace the `\r` instead?

analytics/analytics.go
```diff
@@ -0,0 +1,258 @@
...
+
+	WriteConsent()
+}
+
+// WriteConsent writes the current consent state to a persistent file
+func WriteConsent() {
+	c := cliTrackingConsent{ConsentGiven: ConsentGiven}
+	json, err := json.Marshal(c)
+	if err != nil {
+		Println("Error: unable to record consent. Exiting.")
```
pavelnikolov: Both this error and the next one are the same. This doesn't help you much with debugging in case of an error.

analytics/analytics.go
```diff
@@ -0,0 +1,258 @@
...
+// Behavior is determined by consent:
+//
+// if consent not given {
+// 	prompt for consent
+// }
+// if consent == true {
+// 	submit analytics
+// }
+func Submit(e Event) (err error) {
+	ReadConsent()
```
pavelnikolov: Read consent doesn't return an error? 🤔

analytics/analytics.go
```diff
@@ -0,0 +1,258 @@
...
+	if !ConsentGiven {
+		return err
```
pavelnikolov, line 234: Isn't error `nil` in this case? You don't assign anything to it. Am I missing something?

analytics/analytics_test.go
```diff
@@ -0,0 +1,67 @@
...
+	var called bool
+	consentPath = newConsentTempfile(t)
+	ConsentGiven = true
+	WriteConsent()
+
+	// Setup
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		called = true
+
+		body, _ := ioutil.ReadAll(r.Body)
```
pavelnikolov: Read the error and fail the test.

analytics/consent_test.go
```diff
@@ -0,0 +1,148 @@
...
+	assert := assert.New(t)
+
+	testCases := []string{
+		"\n",
+		"n\n",
+		"N\n",
+		"OESNAOSENAO\n",
+		"😈\n",
+	}
+	for _, tc := range testCases {
```
pavelnikolov, line 95: These tests wouldn't work in parallel because you use global variables. This is not immediately obvious here. I can come in one year and add `t.Parallel()` and break this. Sometimes. So better avoid global vars.

pavelnikolov (approved):

analytics/analytics.go
auxesis: Fixed from 40d8ac30393e5f74efdece620d0e8e0af78469d2 onwards.

Makefile
auxesis: Unfortunately no:

```
go build -ldflags "-X 'HeapAppID=892134159'" -o bin/section section.go
# command-line-arguments
/usr/local/Cellar/go/1.14.6/libexec/pkg/tool/darwin_amd64/link: -X flag requires argument of the form importpath.name=value
make: *** [build] Error 2
```

analytics/analytics.go
auxesis: Happy to, but `go fmt` rewrites it to what's committed.

analytics/analytics.go
auxesis: Almost! 

The `HeapAppID` in the source code is the ID of our development environment. 

During the build, we override it with the production ID. 

The rationale here is to keep analytics generated during dev/test separate to production usage.

pavelnikolov:
go fmt doesnt rewrite empty lines in imports

analytics/analytics_test.go
auxesis: Fixed in 9063fd03d10f0ea908f26bee701143a0a9ecc1ed.

analytics/analytics.go
auxesis: Added a comment to this line in 8069d33307a83bb57987606675aa8597f4ef631d to make it clearer what's happening.

analytics/analytics.go
auxesis: Good call, I refactored in ab7d68f1427e248da09fdfefc19523ca74eb5211 to handle the error in one place.

analytics/analytics.go
auxesis: Yep, this is totally fine behavior.

If there is no consent given, `Submit` should return immediately with a `nil` error.

analytics/analytics.go
auxesis: Fixed in e96d5d3d756311a7dfcc281fc6fc1a361a2162f2.

auxesis:
> go fmt doesnt rewrite empty lines in imports

Looks like I was using an older version of fatih/vim-go that was rolling all the imports up. 

I upgraded vim-go, reformatted the imports, and committed the changes in e96d5d3d756311a7dfcc281fc6fc1a361a2162f2.

analytics/analytics.go
auxesis: It now does as of 8a36bbf.

analytics/analytics.go
auxesis: Yep good pickup, have fixed this in 94c2b35d4951ecdad55632fb6053d5bee319732a.

analytics/consent_test.go
auxesis: I understand. 

In this particular case, given this is a simple CLI tool – not a highly concurrent server – I think it's a reasonable maintainability tradeoff.

PR #6, opened 2020-09-16
Add "apps" subcommands

auxesis:
New (sub)commands: 

- `section accounts` for listing accounts
- `section apps list` for listing apps under an account
- `section apps info` for showing detailed info about an app

### DIFF
--- a/api/account.go
+++ b/api/account.go
@@ -1,0 +1,47 @@
+package api
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"net/url"
+)
+
+// Application represents an application deployed on Section
+type Application struct {
+	ID              int    `json:"id"`
+	Href            string `json:"href"`
+	ApplicationName string `json:"application_name"`
+}
+
+// Applications returns a list of applications on a given account.
+func Applications(accountID int) (as []Application, err error) {
+	u, err := url.Parse(BaseURL)
+	if err != nil {
+		log.Fatal(err)
+	}
+	u.Path += fmt.Sprintf("/account/%d/application", accountID)
+
+	resp, err := request(http.MethodGet, u.String(), nil)
+	if err != nil {
+		return as, err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return as, fmt.Errorf("request failed with status %s", resp.Status)
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return as, err
+	}
+
+	err = json.Unmarshal(body, &as)
+	if err != nil {
+		return as, err
+	}
+	return as, err
+}
--- a/api/account.go
+++ b/api/account.go
@@ -6,7 +6,6 @@ import (
 	"io/ioutil"
 	"log"
 	"net/http"
-	"net/url"
 	"sort"
 )
 
@@ -22,7 +21,7 @@ type Account struct {
 
 // Accounts returns a list of account the current user has access to.
 func Accounts() (as []Account, err error) {
-	u, err := url.Parse(BaseURL)
+	u, err := BaseURL()
 	if err != nil {
 		log.Fatal(err)
 	}
--- a/api/account.go
+++ b/api/account.go
@@ -7,22 +7,26 @@ import (
 	"log"
 	"net/http"
 	"net/url"
+	"sort"
 )
 
-// Application represents an application deployed on Section
-type Application struct {
-	ID              int    `json:"id"`
-	Href            string `json:"href"`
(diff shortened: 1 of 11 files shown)


api/api.go
```diff
@@ -0,0 +1,53 @@
...
+	"os/user"
+	"path/filepath"
+	"time"
+
+	"github.com/jdxcode/netrc"
+)
+
+var (
+	// BaseURL is the root of the Section API
+	BaseURL = "https://aperture.section.io/api/v1"
```
jstangroome: Would be handy to override via environment variable, especially for testing. One convention elsewhere has been `APERTURE_URI_PREFIX=https://aperture.section.io` but open to other ideas.

api/applications.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	envs, err := ApplicationEnvironments(accountID, applicationID)
+	if err != nil {
+		return a, err
+	}
+	a.Environments = envs
+
+	for i, e := range a.Environments {
+		stack, err := ApplicationEnvironmentStack(accountID, applicationID, e.EnvironmentName)
```
jstangroome, line 78: Would be nice to better understand some use cases if we can provide a native API server side to avoid the "N+1 query" problem.

auxesis: 💯 

I've been writing down where I'm finding the API lacking as I go. 

This particular part of the API is a bit cludgy.

We should definitely go back and refine some of the existing API endpoints based on what I'm learning here.

api/api.go
auxesis: Yeah that's a good call. 

Right now I am changing this variable inside the test suite, but I can see use cases where you might want to override it from the CLI.

api/api.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil {
+		return u, p, err
+	}
+	u = n.Machine("aperture.section.io").Get("login")
+	p = n.Machine("aperture.section.io").Get("password")
+	return u, p, err
+}
+
+func request(method string, url string, body io.Reader) (resp *http.Response, err error) {
+	client := &http.Client{
```
jstangroome, line 42: Can we set the User-Agent header to at least specify section-cli, and ideally contain the binary version injected by the CI process? E.g.:

- [set the header](https://github.com/section-io/varnish-cli-bridge/blob/5fbc9bb62c73db6e2bf716e6b70f7d48a19d08b0/api.go#L33)
- [define the value from build-time values](https://github.com/section-io/varnish-cli-bridge/blob/5fbc9bb62c73db6e2bf716e6b70f7d48a19d08b0/main.go#L81)
- [inject the build-time values](https://github.com/section-io/varnish-cli-bridge/blob/7890759da5d61bbba69650c7602ac222fbc03dab/.travis.yml#L10)

jstangroome (approved):

api/api.go
auxesis: @jstangroome good call, those are great suggestions.

I'll add them to this PR before I merge.

api/account.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+// Account represents an account on Section
+type Account struct {
+	ID          int    `json:"id"`
+	Href        string `json:"href"`
+	AccountName string `json:"account_name"`
+	IsAdmin     bool   `json:"is_admin"`
+	BillingUser int    `json:"billing_user"`
+	Requires2FA bool   `json:"requires_2fa"`
```
pavelnikolov, line 19: Shouldn't 2FA be used only on the server side?

auxesis: Yep, but this is just repeating attributes that the API exposes.

api/applications.go
pavelnikolov: You should not touch the existing API endpoinnts. There is a new GraphQL schema. You can add them there.

api/account.go
pavelnikolov: The GraphQL schema should not expose fields that we do not need.

pavelnikolov: Also, if the input JSON has more fields than the struct, that is fine, they will be ignored.

api/applications.go
pavelnikolov: We should avoid using the old APIs from now on and only use the GraphQL API if we want to decommission Sails.

api/api.go
auxesis: Fixed in e7fbf2e. 

Can be set by either: 

- Specifying the `--section-api-prefix` flag
- Specifying the `SECTION_API_PREFIX` environment variable

api/api.go
auxesis: Fixed in e7e9968.

api/applications.go
auxesis: Good call. 

We'll need to come back and rework the API client to work with the new GraphQL API after launch.

PR #7, opened 2020-09-21
Add login subcommand

auxesis:
Works like this: 

section login screencast


api/auth/auth.go
```diff
@@ -0,0 +1,154 @@
...
+
+	restoreTerminal()
+	return m, u, p, err
+}
+
+// WriteCredential saves Section API credentials to disk
+func WriteCredential(machine, username, password string) (err error) {
+	_, err = os.Stat(CredentialPath)
+	if os.IsNotExist(err) {
+		file, err := os.Create(CredentialPath)
```
jstangroome, line 116: Do you also need to recursively `mkdir` if the directory structure is absent?

auxesis: @jstangroome ugh yes, that's a good pickup, thank you!

api/auth/auth.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	u := n.Machine("aperture.section.io").Get("login")
+	p := n.Machine("aperture.section.io").Get("password")
+	return (len(u) > 0 && len(p) > 0)
+}
+
+// PromptForCredential interactively prompts the user for a credential to authenticate to the Section API
+func PromptForCredential() (m, u, p string, err error) {
+	m = "aperture.section.io"
```
jstangroome: Shouldn't all these hard-coded `aperture.section.io` values come instead from the host component of the `SECTION_API_PREFIX`/`--section-api-prefix` overrides, so that credentials for one endpoint aren't sent to another?

api/auth/auth.go
auxesis: Fixed in ecedb5db81ab0187a7ed6c5bd11341bf8b4ecde2.

api/auth/auth.go
auxesis: That's a good point I had not considered.

We should set the machine name in the netrc based on whatever value is set for `api.PrefixURI`.

We should also display what endpoint we're asking for credentials for when showing the prompt. 

I'll make those changes now.

mwickman (approved):

api/auth/auth.go
auxesis: Fixed in 3132a77016ed11b7010b4e298a08f6c83a54876b.

PR #11, opened 2020-09-25
Bug fixes from demo walkthrough

auxesis:
- Make URL handling work (kept appending paths onto URLs instead of setting them new every time)
- Add some more informative debugging messages when running the `deploy` subcommand
- Refactor filelist generation, add tests, and make it more robust

### DIFF
--- a/commands/deploy.go
+++ b/commands/deploy.go
@@ -46,6 +46,7 @@ func (c *DeployCmd) Run() (err error) {
 		return fmt.Errorf("failed to walk path: %v", err)
 	}
 	if c.Debug {
+		fmt.Println("Archiving files:")
 		for _, file := range files {
 			fmt.Println(file)
 		}
--- a/commands/deploy.go
+++ b/commands/deploy.go
@@ -19,7 +19,7 @@ const MaxFileSize = 1073741824 // 1GB
 // DeployCmd handles deploying an app to Section.
 type DeployCmd struct {
 	Debug     bool
-	Directory string `default:"./"`
+	Directory string `default:"."`
 	ServerURL string `default:"https://aperture.section.io/new/code_upload/v1"`
 }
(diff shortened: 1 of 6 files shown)


auxesis:
Test is failing but only intermittently, and only in GitHub Actions 😦 . 

Should merge cleanly

commands/deploy.go
```diff
@@ -19,6 +19,7 @@ const MaxFileSize = 1073741824 // 1GB
 // DeployCmd handles deploying an app to Section.
 type DeployCmd struct {
 	Debug     bool
+	Directory string `default:"./"`
```
pavelnikolov: Maybe this could work just with a `.`.

commands/deploy_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+func TestCommandsDeployBuildFilelistErrorsOnNonExistentDirectory(t *testing.T) {
+	assert := assert.New(t)
+
+	// Setup
+	testCases := []string{
+		filepath.Join("testdata", "deploy", "non-existent-tree"),
+		filepath.Join("testdata", "deploy", "file"),
+	}
+	ignores := []string{}
```
pavelnikolov: (nitpick) I prefer the default values:
`var ignores []string`

pavelnikolov (approved):

commands/deploy_test.go
auxesis: Good pickup, I prefer that syntax too. 

Fixed in 9a32807.

commands/deploy.go
auxesis: Yep good call. 

I did some testing, and this works fine. 

Changed in 871ba3c55c15bf5e0cc0db93d6ee633335d1fe8c.

PR #13, opened 2020-09-28
Clean up and simplify analytics package

auxesis:
- Explicitly pass stdin/stdout around, to eliminate globals. 
- Simplify build commands, now we have dropped custom `Printf`, and can use `fmt.Fprintf`


analytics/analytics.go
```diff
@@ -125,9 +115,9 @@ func IsConsentRecorded() (rec bool) {
 }
 
 // ReadConsent finds if consent has been given, either from file or by prompt.
-func ReadConsent() (c bool, err error) {
+func ReadConsent(in io.Reader, out io.Writer) (c bool, err error) {
```
pavelnikolov, line 118: That's much better.

analytics/consent_test.go
```diff
@@ -49,13 +49,13 @@ func TestConsentPromptsForConsentIfConsentNotRecorded(t *testing.T) {
 	consentPath = newConsentTempfile(t)
 
 	var outbuf bytes.Buffer
-	out = &outbuf
+	out := &outbuf
 	var inbuf bytes.Buffer
 	inbuf.Write([]byte("\n"))
-	in = &inbuf
+	in := &inbuf
```
pavelnikolov: Do you really need these two variables (`in` and `out`)?

pavelnikolov (approved):

analytics/consent_test.go
auxesis: @pavelnikolov good call – I have refactored them out in 45d1c59.

PR #14, opened 2020-09-29
Add authentication to deploy command

auxesis:
- Ensure HTTP requests to the upload service include basic auth. 
- Treat URLs as a `url.URL`, to make extracting info easier, and keep consistent with the rest of the code base. 
- Include some extra debugging info on errors.

### DIFF
--- a/commands/deploy.go
+++ b/commands/deploy.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


commands/deploy.go
```diff
@@ -57,22 +67,29 @@ func (c *DeployCmd) Run() (err error) {
...
 	client := &http.Client{
 		Timeout: 30 * time.Second,
 	}
-	req, err := http.NewRequest(http.MethodPost, c.ServerURL, tempFile)
+	req, err := http.NewRequest(http.MethodPost, c.ServerURL.String(), tempFile)
 	if err != nil {
 		return fmt.Errorf("failed to create upload URL: %v", err)
 	}
+
+	username, password, err := auth.GetCredential(c.ServerURL.Host)
```
pavelnikolov, line 80: Where is the password stored?

pavelnikolov (approved):

commands/deploy.go
auxesis: This was introduced in #7. 

Password is [pulled from](https://github.com/section/sectionctl/blob/0d157b5fb78eb2ddc75769aeb1a8568cae3f70fe/api/auth/auth.go#L135-L147) a netrc in `~/.config/section`.

PR #15, opened 2020-09-30
Add deploy subcommand tests

auxesis:
Also fixes a bug where the tarball isn't actually uploaded.


commands/deploy_test.go
```diff
@@ -48,3 +55,51 @@ func TestCommandsDeployBuildFilelistErrorsOnNonExistentDirectory(t *testing.T) {
...
+	c := DeployCmd{Directory: dir, ServerURL: url}
+	err = c.Run()
+
+	assert.NoError(err)
+	assert.True(req.called)
+	assert.Equal(username, req.username)
+	assert.Equal(password, req.password)
+	assert.NotZero(len(req.body))
+
+	mime, err := mimetype.DetectReader(bytes.NewReader(req.body))
```
jstangroome: You could just check if the first two bytes are 0x1f and 0x8b, i.e. the gzip header, and save the ongoing maintenance of bringing in another third party dependency. Check the third byte is 0x08 if deflate compression is important.

jstangroome (approved):

commands/deploy_test.go
auxesis: That's quite a pro-tip! I'll change that now.

PR #17, opened 2020-10-13
Submit metadata and tarball as a multipart form

auxesis:
Make the file upload a multipart form, so `sectionctl` can send the:

- File
- Account + app IDs

### DIFF
--- a/commands/deploy.go
+++ b/commands/deploy.go
@@ -8,10 +8,12 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"mime/multipart"
 	"net/http"
 	"net/url"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
@@ -85,12 +87,9 @@ func (c *DeployCmd) Run() (err error) {
 
 	fmt.Printf("Pushing %d bytes...\n", stat.Size())
 
-	client := &http.Client{
-		Timeout: 30 * time.Second,
-	}
-	req, err := http.NewRequest(http.MethodPost, c.ServerURL.String(), tempFile)
+	req, err := newFileUploadRequest(c, tempFile)
 	if err != nil {
-		return fmt.Errorf("failed to create upload URL: %v", err)
+		return fmt.Errorf("unable to build file upload: %s", err)
 	}
 
 	username, password, err := auth.GetCredential(c.ServerURL.Host)
@@ -99,6 +98,9 @@ func (c *DeployCmd) Run() (err error) {
 	}
 	req.SetBasicAuth(username, password)
 
+	client := &http.Client{
+		Timeout: 30 * time.Second,
+	}
 	resp, err := client.Do(req)
 	if err != nil {
 		return fmt.Errorf("upload request failed: %v", err)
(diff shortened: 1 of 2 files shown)


auxesis:
For reference:

- I [used this gist](https://gist.github.com/mattetti/5914158/f4d1393d83ebedc682a3c8e7bdc6b49670083b84) as a reference for doing the upload in `sectionctl`.
- [This comment](https://gist.github.com/mattetti/5914158/f4d1393d83ebedc682a3c8e7bdc6b49670083b84#gistcomment-1237868) in the above gist to set the content type correctly.
- I [used this StackOverflow answer](https://stackoverflow.com/a/51345650) as a reference for handling the upload in the tests.
- There are more [memory efficient ways](https://gist.github.com/mattetti/5914158/f4d1393d83ebedc682a3c8e7bdc6b49670083b84#gistcomment-1985369) of handling the file on both the client and server side.

commands/deploy.go
```diff
@@ -99,6 +98,9 @@ func (c *DeployCmd) Run() (err error) {
 	}
 	req.SetBasicAuth(username, password)
 
+	client := &http.Client{
+		Timeout: 30 * time.Second,
```
pavelnikolov: Unrelated to this PR but if the content is too big, 30 seconds might not be enough to deploy it. I wonder if this needs to be configurable with a 30s default.

commands/deploy.go
```diff
@@ -250,3 +252,45 @@ func triggerUpdate(accountID, appID int, payloadID, serviceURL string, c *http.C
...
+
+// newFileUploadRequest builds a HTTP request for uploading an app and the account + app it belongs to
+func newFileUploadRequest(c *DeployCmd, f *os.File) (r *http.Request, err error) {
+	contents, err := ioutil.ReadAll(f)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+
+	body := new(bytes.Buffer)
```
pavelnikolov: (nitpick) I prefer `var body bytes.Buffer`. No need for `new` here. Again use default value as per Rob Pike's proverbs.

pavelnikolov (approved):

commands/deploy.go
```diff
@@ -250,3 +252,45 @@ func triggerUpdate(accountID, appID int, payloadID, serviceURL string, c *http.C
 	}
 	return nil
 }
+
+// newFileUploadRequest builds a HTTP request for uploading an app and the account + app it belongs to
+func newFileUploadRequest(c *DeployCmd, f *os.File) (r *http.Request, err error) {
+	contents, err := ioutil.ReadAll(f)
```
jstangroome, line 259: Has this just re-introduced that the cli needs to fit the entire payload into ram?

jstangroome: This is likely a better pattern: https://gist.github.com/ZenGround0/49e4a1aa126736f966a1dfdcb84abdae

commands/deploy.go
auxesis: I've been thinking about this as well. 

I think the timeout should be 

- 60s by default
- configurable on the command line

commands/deploy.go
auxesis: Good call, this was a copypasta from a gist 😂

commands/deploy.go
auxesis: 💯  I'm working on an optimization for this now.

commands/deploy.go
auxesis: So fun little thing here: 

The `new` allocates zero memory to the variable. 

The more idiomatic `var body *bytes.Buffer` causes a panic in `multipart.NewWriter(body)` because it's a nil pointer dereference.

I'm gonna change this line to be `body := &bytes.Buffer{}` so the code style stays consistent by eliminating the `new` and still not cause `multipart.NewWriter()` to panic.

commands/deploy.go
auxesis: Fixed in 0338747.

commands/deploy.go
pavelnikolov: No, you are supposed to do this:

```
var body bytes.Buffer
```

then use `&body` as a parameter to the func.

commands/deploy.go
```diff
@@ -261,7 +262,7 @@ func newFileUploadRequest(c *DeployCmd, f *os.File) (r *http.Request, err error)
 	}
 	defer f.Close()
 
-	body := new(bytes.Buffer)
+	body := &bytes.Buffer{}
```
pavelnikolov: `var body bytes.Buffer`

commands/deploy.go
```diff
@@ -261,7 +262,7 @@ func newFileUploadRequest(c *DeployCmd, f *os.File) (r *http.Request, err error)
...
 	writer := multipart.NewWriter(body)
```
pavelnikolov: `writer := multipart.NewWriter(&body)`

auxesis: I was getting some weird behavior before when I was trying this, but it appears to be working fine. 

Fixed in 1bb4712b1a1f3a73dcb191c03340c6fcc9d10ac8.

commands/deploy.go
auxesis: I'm going to create a separate PR for the memory optimizations, in the `make-file-upload-memory-efficient` branch.

Right now, we just want `sectionctl` to work. Then we can focus on making it right.

PR #18, opened 2020-10-15
Make file upload memory efficient

auxesis:
Follow up to #17. 

The intent of this PR is to make the upload of the tarball memory efficient, by streaming it off disk, instead of reading it into memory.


auxesis:
Closing this as WONTFIX until we get bug reports from users about it.

PR #19, opened 2020-10-20
Improve deploy subcommand debug statements

auxesis:
- Prefix all existing debug messages with `[debug]`, to improve readability
- Make it clearer to the user when errors are being wrapped
- Display bytes transferred in a unit that's more readable to the user
- Emit the request URLs explicitly every time the request is being made, instead of just printing user input
- Print the path to the temporary tarball

### DIFF
--- a/commands/deploy.go
+++ b/commands/deploy.go
@@ -69,6 +69,9 @@ func (c *DeployCmd) Run() (err error) {
 	if err != nil {
 		return fmt.Errorf("failed to pack files: %v", err)
 	}
+	if c.Debug {
+		fmt.Println("[debug] Temporary tarball path:", tempFile.Name())
+	}
 	stat, err := tempFile.Stat()
 	if err != nil {
 		return fmt.Errorf(fmt.Sprintf("Could not get stat for file '%s', got error '%s'", tempFile.Name(), err.Error()))
--- a/commands/deploy.go
+++ b/commands/deploy.go
@@ -119,8 +119,8 @@ func (c *DeployCmd) Run() (err error) {
 	if err != nil {
 		return fmt.Errorf("failed to decode response %v", err)
 	}
-	svcURL := c.ApertureURL + fmt.Sprintf(c.EnvUpdatePathFmt, c.AccountID, c.AppID, "production")
-	err = triggerUpdate(c.AccountID, c.AppID, response.PayloadID, svcURL, client)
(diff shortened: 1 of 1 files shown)


commands/deploy.go
```diff
@@ -86,7 +85,7 @@ func (c *DeployCmd) Run() (err error) {
 		return fmt.Errorf("unable to seek to beginning of tarball: %s", err)
 	}
 
-	fmt.Printf("Pushing %d bytes...\n", stat.Size())
+	fmt.Printf("Pushing %dMB...\n", stat.Size()/1024)
```
pavelnikolov, line 88: Is `Size()` returning kilobytes?

pavelnikolov (approved):

commands/deploy.go
```diff
@@ -42,16 +42,12 @@ type UploadResponse struct {
...
 func (c *DeployCmd) Run() (err error) {
-	if c.Debug {
-		fmt.Println("Server URL:", c.ServerURL.String())
-	}
-
 	ignores := []string{".lint/", ".git/"}
 	files, err := BuildFilelist(c.Directory, ignores)
 	if c.Debug {
-		fmt.Println("Archiving files:")
+		fmt.Println("[debug] Archiving files:")
```
jstangroome, line 48: Consider the [hashicorp logging package](https://github.com/hashicorp/logutils) that we use in many other Go projects. It can filter the log output just by the presence of prefixes like `[debug]` in the written string without needing to litter code with `if loglevel {}` blocks.

commands/deploy.go
auxesis: I believe it's bytes. 

[godoc says](https://golang.org/pkg/os/#FileInfo):

```
Size() int64        // length in bytes for regular files; system-dependent for others
```

commands/deploy.go
auxesis: @jstangroome yep good call. 

Now our logging is becoming more verbose, we're gonna better sensible defaults for logging.

Will put up another PR to improve logging.

commands/deploy.go
pavelnikolov: Then you need to divide by `1024` twice in order to get `MB`. Your message is wrong.

auxesis: 😂 I have never felt more post-technical in my life.

PR #26, opened 2020-10-21
Increase the deploy timeout to 5 mins

auxesis:
Uploads can be big, so we need to wait a while for them to succeed.


pavelnikolov (approved):

commands/deploy.go
```diff
@@ -32,7 +32,7 @@ type DeployCmd struct {
 	ServerURL        *url.URL      `default:"https://aperture.section.io/new/code_upload/v1/upload"`
 	ApertureURL      string        `default:"https://aperture.section.io/api/v1"`
 	EnvUpdatePathFmt string        `default:"/account/%d/application/%d/environment/%s/update"`
-	Timeout          time.Duration `default:"60s"`
+	Timeout          time.Duration `default:"300s"`
```
jstangroome, line 35: The problem may be that we have a total elapsed time limit, but should have an idle timeout?

pavelnikolov: How do you set the iddle time limit?

PR #27, opened 2020-10-21
Verify that what's being deployed looks like a Node.js app

auxesis:
Tests that: 

- `node_modules/` is a directory in the target app directory
- `package.json` is a file in the target app directory

### DIFF
--- a/commands/deploy.go
+++ b/commands/deploy.go
@@ -60,6 +60,16 @@ func (c *DeployCmd) Run() (err error) {
 	}
 	fmt.Printf("Packaging app in: %s\n", dir)
 
+	errs := IsValidNodeApp(dir)
+	if len(errs) > 0 {
+		var se []string
+		for _, err := range errs {
+			se = append(se, fmt.Sprintf("- %s", err))
+		}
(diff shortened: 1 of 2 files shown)


commands/deploy.go
```diff
@@ -60,6 +60,16 @@ func (c *DeployCmd) Run() (err error) {
...
 
+	errs := IsValidNodeApp(dir)
+	if len(errs) > 0 {
+		var se []string
+		for _, err := range errs {
+			se = append(se, fmt.Sprintf("- %s", err))
+		}
+		errstr := strings.Join(se, "\n")
+		return fmt.Errorf("not a valid Node.js app: \n\n%s", errstr)
+	}
```
pavelnikolov, line 71: I think that we should print only one error at a time. The user will fix only one error at a time anyway.

pavelnikolov (approved):

commands/deploy.go
auxesis: I originally thought this too. 

On reflection, it can be frustrating being drip-fed errors. 

I want to give this a try, and see what users have to say about it.

PR #30, opened 2020-10-23
Fix tar creation

manibatra:
We were not writing the tar header for directories which was causing issues when trying to untar the tar.gz in the configmap watcher. This PR fixes the issue.

### DIFF
--- a/commands/deploy.go
+++ b/commands/deploy.go
@@ -183,9 +183,6 @@ func BuildFilelist(dir string, ignores []string) (files []string, err error) {
 	}
 
 	err = filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
-		if info.IsDir() {
-			return nil
-		}
 		for _, i := range ignores {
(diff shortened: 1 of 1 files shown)


auxesis (approved):
Nice work getting this fix up. LGTM.

commands/deploy.go
```diff
@@ -231,21 +228,25 @@ func addFileToTarWriter(filePath string, tarWriter *tar.Writer) error {
...
-		Mode:    int64(stat.Mode()),
-		ModTime: stat.ModTime(),
+	header, err := tar.FileInfoHeader(stat, filePath)
+	if err != nil {
+		return err
 	}
 
+	// must provide real name
+	// (see https://golang.org/src/archive/tar/common.go?#L626)
+	header.Name = filepath.ToSlash(filePath)
```
jstangroome, line 238: `filepath` package versus `filePath` variable looks like a recipe for future maintenance headaches. Consider not depending on case as the only difference.

PR #35, opened 2020-10-27
Create symlinks in the upload tarball made by sectionctl deploy

auxesis:
So complex link structures under `node_modules/` work.

### DIFF
--- a/commands/deploy.go
+++ b/commands/deploy.go
@@ -244,7 +244,7 @@ func addFileToTarWriter(filePath string, tarWriter *tar.Writer, prefix string) e
 	if err != nil {
 		return err
 	}
-	if stat.Mode()&os.ModeSymlink != 0 {
+	if stat.Mode()&os.ModeSymlink == os.ModeSymlink {
 		link, err := os.Readlink(filePath)
 		if err != nil {
 			return err
--- a/commands/deploy.go
+++ b/commands/deploy.go
@@ -234,7 +234,7 @@ func addFileToTarWriter(filePath string, tarWriter *tar.Writer, prefix string) e
 	}
 	defer file.Close()
 
-	stat, err := file.Stat()
+	stat, err := os.Lstat(filePath)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


commands/deploy.go
```diff
@@ -244,6 +244,13 @@ func addFileToTarWriter(filePath string, tarWriter *tar.Writer, prefix string) e
 	if err != nil {
 		return err
 	}
+	if stat.Mode()&os.ModeSymlink != 0 {
```
jstangroome: Technically more correct where the mask is multibit and good habit so you can ignore how many bits the flag/mask is.
```suggestion
	if stat.Mode()&os.ModeSymlink == os.ModeSymlink {
```

commands/deploy.go
```diff
@@ -254,7 +261,7 @@ func addFileToTarWriter(filePath string, tarWriter *tar.Writer, prefix string) e
 		return fmt.Errorf(fmt.Sprintf("Could not write header for file '%s', got error '%s'", baseFilePath, err.Error()))
 	}
 
-	if !stat.IsDir() {
+	if !stat.IsDir() && stat.Mode()&os.ModeSymlink == 0 {
```
jstangroome: Again, per multi-bit comment above.
```suggestion
	if !stat.IsDir() && stat.Mode()&os.ModeSymlink != os.ModeSymlink {
```

jstangroome (approved):

jstangroome:
At what point of having reimplemented much of tar in our Go code do we instead insist that the user has `tar` on their PATH and we then shell out? :)

auxesis:
To be honest, given the scope of what we're trying to do here with compressed archives, shelling out to `tar` might actually be more maintainable. 

One potential problem I can see is the differences between GNU and BSD tar – we'd need to use a subset of features and syntax that both support.

jstangroome:
> One potential problem I can see is the differences between GNU and BSD tar – we'd need to use a subset of features and syntax that both support.

Looks like Windows has the BSD flavour of tar:
https://techcommunity.microsoft.com/t5/containers/tar-and-curl-come-to-windows/ba-p/382409

PR #37, opened 2020-10-28
Make deploy flags required, and set no defaults

auxesis:
User must specify the flags, so they are explicitly deploying a specific app.

### DIFF
--- a/commands/deploy.go
+++ b/commands/deploy.go
@@ -26,8 +26,8 @@ const MaxFileSize = 1073741824 // 1GB
 
 // DeployCmd handles deploying an app to Section.
 type DeployCmd struct {
-	AccountID        int `default:"4322"`  // harc-coded until authentication is implemented
-	AppID            int `default:"65443"` // hard-coded for now until authentication is implmented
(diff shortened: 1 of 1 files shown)


pavelnikolov (approved):

commands/deploy.go
```diff
@@ -26,8 +26,8 @@ const MaxFileSize = 1073741824 // 1GB
 
 // DeployCmd handles deploying an app to Section.
 type DeployCmd struct {
-	AccountID        int `default:"4322"`  // harc-coded until authentication is implemented
-	AppID            int `default:"65443"` // hard-coded for now until authentication is implmented
+	AccountID        int `required`
```
jstangroome, line 29: Tools like the GCloud SDK allow you to specify default values in the user's configuration file to avoid passing the project-id every time. Similar to AWS CLI storing the default region. Should we consider this?

auxesis: Yep, and I even have a half written PR for setting defaults. 

Need to get that published and released in this week!

PR #43, opened 2020-11-10
Add ability to set credentials via environment variables

### LINKED ISSUE #33: Allow specifying username/password credential as environment variable
This will make it easier to use `sectionctl` in CD pipelines. 

Environment variables could be `SECTION_USERNAME` and `SECTION_PASSWORD`, and eventually `SECTION_TOKEN` once we support standalone tokens.

auxesis:
Allows credentials to be set via the `SECTION_USERNAME` and `SECTION_TOKEN` environment variables. 

Closes #33.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,10 +26,10 @@ sectionctl install-completions
 
 ### Using `sectionctl` in CI/CD
 
-You can set credentials via the `SECTION_USERNAME` and `SECTION_PASSWORD` environment variables:
+You can set credentials via the `SECTION_USERNAME` and `SECTION_TOKEN` environment variables:
 
-```
-SECTION_USERNAME=me@domain.example SECTION_PASSWORD=s3cr3t sectionctl accounts list
+``` bash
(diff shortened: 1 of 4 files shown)


manibatra (approved):

README.md
```diff
@@ -24,6 +24,14 @@ Install bash shell completions with:
 sectionctl install-completions
 ```
 
+### Using `sectionctl` in CI/CD
+
+You can set credentials via the `SECTION_USERNAME` and `SECTION_TOKEN` environment variables:
+
+``` bash
+SECTION_USERNAME=me@domain.example SECTION_TOKEN=s3cr3t sectionctl accounts list
```
jstangroome, line 32: Often interactive bash shells have `HISTCONTROL=ignorespace` [or equivalent](https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html#index-HISTCONTROL) defined in the default profile, which means that using a leading space on a command like this will prevent it from being logged in the bash history file with the token. Not sure if or how to communicate this to a user.

jstangroome (approved):

PR #44, opened 2020-11-11
Refactor HTTP requests made in `sectionctl deploy` into the API client

### LINKED ISSUE #23: Make deploy subcommand use the api package
Currently it makes all its HTTP requests itself. 

It should use a dedicated function `github.com/section/sectionctl/api` package.

auxesis:
Closes #23.

Also: 

- Adds support for specifying headers on API requests if needed.
- Adds debug output for all API requests if the `--debug` flag or `DEBUG` environment variable is set.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -20,6 +20,8 @@ var (
 	Username string
 	// Token is the token for authenticating to the Section API
 	Token string
+	// Debug toggles whether extra information is emitted from requests/responses
+	Debug bool
 )
 
 // BaseURL returns a URL for building requests on
@@ -57,6 +59,14 @@ func request(method string, u url.URL, body io.Reader, headers ...map[string][]s
 	}
 	req.SetBasicAuth(Username, Token)
 
+	if Debug {
+		fmt.Println("[DEBUG] Request URL:", req.URL)
+		for k, vs := range req.Header {
+			for _, v := range vs {
+				fmt.Printf("[DEBUG] Header: %s: %v\n", k, v)
+			}
+		}
+	}
 	resp, err = client.Do(req)
 	if err != nil {
 		return resp, err
--- a/api/api.go
+++ b/api/api.go
@@ -31,7 +31,7 @@ func BaseURL() (u url.URL) {
 	return u
 }
 
-func request(method string, u url.URL, body io.Reader, headers ...map[string][]string) (resp *http.Response, err error) {
+func request(method string, u url.URL, body io.Reader, headers ...http.Header) (resp *http.Response, err error) {
 	client := &http.Client{
 		Timeout: timeout,
 	}
--- a/api/api.go
+++ b/api/api.go
@@ -31,6 +31,9 @@ func BaseURL() (u url.URL) {
 	return u
 }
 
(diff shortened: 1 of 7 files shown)


api/api.go
```diff
@@ -43,15 +45,30 @@ func request(method string, u url.URL, body io.Reader) (resp *http.Response, err
 	req.Header.Set("User-Agent", ua)
 	req.Header.Add("Content-Type", "application/json")
 	req.Header.Add("Accept", "application/json")
+	if len(headers) == 1 {
```
mwickman: Should this be `>= 1` ? Seems a bit odd to use a `map[string][]string` if we are only expecting a single header value. Maybe we should accept `http.Header` as the arg instead?

commands/deploy.go
```diff
@@ -158,13 +162,15 @@ func (c *DeployCmd) Run() (err error) {
...
 	s.Start()
-	serviceURL := c.ApertureURL + fmt.Sprintf(c.EnvUpdatePathFmt, c.AccountID, c.AppID, "production")
-	err = triggerUpdate(c, response.PayloadID, serviceURL, client)
+	up := api.EnvironmentUpdateCommand{
+		Op: "replace",
+		Value: PayloadValue{
+			ID: response.PayloadID,
+		},
+	}
+	err = api.ApplicationEnvironmentModuleUpdate(c.AccountID, c.AppID, "production", "nodejs/.section-external-source.json", up)
```
mwickman, line 171: Not asking to change here but just wondering, is "production" akin to the branch name? Does this work for one of the "staging" environments? Maybe we never make those anyways with the NodeJS app flow, but wondering if there is future work we should record.

commands/deploy_test.go
```diff
@@ -210,6 +212,7 @@ func TestCommandsDeployUploadsTarball(t *testing.T) {
 		body      []byte
 		accountID int
 		file      []byte
+		headers   map[string]string
```
mwickman: I still lean towards us re-using the http.Header type in places like this, since it has certain useful methods already attached and shows the intent very clearly. Not a pr-breaker though.

mwickman (approved):
LGTM, thanks for the refactor!

api/api.go
auxesis: Because we're using a variadic function (and thus the caller could pass in as many `map[string][]string` arguments as they like), I've defined this header adding behavior to only run if there's a single collection of headers passed. 

That said, we could probably iterate over all the header arguments and merge them together. 

Will refactor this some more.

commands/deploy.go
auxesis: That's right, `production` is the branch name, so this will only work for that environment. 

I have a todo to add support for multiple environments, which I have recorded as #45.

commands/deploy_test.go
auxesis: Good call, I have fixed this is a61f6c6.

api/api.go
auxesis: As of da55706, the behavior is to merge all the headers passed into a single entity on the HTTP request.

PR #46, opened 2020-11-12
Make deploy environment configurable

### LINKED ISSUE #45: Add support for multiple environments in `sectionctl deploy`
Currently the `deploy` subcommand is hard coded to only work with the `production` environment. 

We should default to `production`, but allow configuration as a flag on the CLI.

auxesis:
What it says on the tin. 

Also adds help text for each flag. 

Also, closes #45.

### DIFF
--- a/analytics/analytics.go
+++ b/analytics/analytics.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"log"
 	"net/http"
 	"os"
 	"os/user"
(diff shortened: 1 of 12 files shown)


pavelnikolov (approved):

jstangroome:
All your examples in the code use the Environment name but IIRC the API will accept name or ID. Can we confirm IDs work from `sectionctl` too?

auxesis:
Can confirm both the `string` and `int` values work.

PR #47, opened 2020-11-12
Do logging through hashicorp/logutils

### LINKED ISSUE #22: Do all logging through hashicorp/logutils

auxesis:
Closes #22.

Puts all debug, warning, and error level output through the `log` package. 

Set the log level based on the `--debug` flag at the beginning of execution. 

Normal user feedback is still being output by `fmt.Print*`. I'm not sure if we should change all of them to log at the `INFO` level. On the one hand, it would be consistent. On the other, there will be a _lot_ of boilerplate output that will likely be confusing to users. 

WDYT @jstangroome?

### DIFF
--- a/analytics/analytics.go
+++ b/analytics/analytics.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"log"
 	"net/http"
 	"os"
 	"os/user"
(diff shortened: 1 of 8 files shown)


mwickman (approved):

jstangroome:
> WDYT @jstangroome?

Leave the fmt.Print without INFO tags.

jstangroome (approved):

PR #51, opened 2020-12-03
Add command to see application status

interludic2000:


auxesis:
As this is the first command to use the new GraphQL API, I'm gonna add some tests to this before I merge.

mwickman:
I tested this against prod and confirmed it works for me! I think we should merge 
image

Issue #25, opened 2020-10-21
Manage sectionctl secrets with go-keyring


auxesis:
https://github.com/zalando/go-keyring

auxesis:
Implemented in #53.

PR #53, opened 2020-12-08
Switch to tokens for authentication

auxesis:
Move away from using username/password to standalone personal access tokens.


auxesis:
Closes #25.

mwickman (approved):

PR #57, opened 2020-12-09
Make analytics async

auxesis:
Pulls analytics off the critical execution path, so `sectionctl` runs faster for users. 

When users have given consent, and analytics events need to be recorded, spawn a separate process to send the event.

The only visible change to users should be that `sectionctl` now runs a lot faster (because a blocking HTTP POST to Heap now happens async in the background).

### DIFF
--- a/commands/analytics.go
+++ b/commands/analytics.go
@@ -1,0 +1,24 @@
+package commands
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/section/sectionctl/analytics"
+)
+
+// AnalyticsCmd handles recording analytics events
+type AnalyticsCmd struct {
+	Event string `required`
+}
+
+// Run executes the command
+func (c *AnalyticsCmd) Run() (err error) {
+	var e analytics.Event
+	err = json.Unmarshal([]byte(c.Event), &e)
+	if err != nil {
+		return fmt.Errorf("unable to parse JSON event: %w", err)
+	}
+	err = analytics.Submit(e)
+	return err
+}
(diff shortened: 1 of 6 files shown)


auxesis:
Also adds a `analytics` subcommand, that accepts a JSON formatted event to send to Heap:

```
sectionctl analytics --event '{"Name":"👋","Properties":{"foo":"🔥"}}'
```

The subcommand is hidden, so users don't get confused when they run `sectionctl help`.

commands/analytics.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/section/sectionctl/analytics"
+)
+
+// AnalyticsCmd handles recording analytics events
+type AnalyticsCmd struct {
```
jstangroome, line 11: Does this, and should this, command appear in `--help` output and tab completion?

jstangroome (approved):

commands/analytics.go
auxesis: It does not (because of the [`hidden` parameter](https://github.com/section/sectionctl/pull/57/files#diff-ad4bfb563a6f3de9223c573150a0a0104111969f3055eace937c1226ad3395c7R27)). 

I don't believe it should show in the `--help` output, because it's an internal function of `sectionctl` that users shouldn't use directly. 

I just checked, and indeed it does show up in tab completion 🤦  I will have to work out a way to exclude it from tab completion.

auxesis: This required an upstream change, which can be found at WillAbides/kongplete#9.

auxesis: As of b214dc7, we're using a fork of kongplete that includes the `hidden` fixes. 

Once the PR is merged upstream, I'll change us back.

PR #58, opened 2020-12-10
Add spinners

auxesis:
Add spinners to:

- `accounts`
- `apps`
- `ps`
- `whoami`

Looks like this:

sectionctl spinners

Also adds a helper that returns a preconfigured spinner, so spinners look consistent across all the subcommands.


commands/accounts.go
```diff
@@ -17,7 +19,19 @@ type AccountsListCmd struct{}
...
+	s := NewSpinner()
+
+	err = auth.Setup(api.PrefixURI.Host)
+	if err != nil {
+		return err
+	}
+
+	s.Suffix = " Looking up accounts..."
+	s.Start()
+	time.Sleep(1 * time.Second)
```
jstangroome: What's with the artificial delay?

auxesis: If there's no delay and the user is on a super fast connection, the status message only flashes on the screen very briefly. 

My hypothesis is that briefly visible status messages cause users confusion. 

I don't yet have any data to back up the hypothesis, nor am I wedded to it. Would love your take @jstangroome.

jstangroome: Options that spring to mind:

1. Instead of the spinner disappearing on completion, it could transition to a completion message, e.g, `.: Looking up accounts...` => `* Looking up accounts... done`, then a flash is moot.

1. Instead of waiting a second before beginning the HTTP request, instead:
    1. Show the spinner
    1. Create a channel (type irrelevant)
    1. Start a new goroutine that internally sleeps, then closes the channel.
    1. Back on your original thread, perform the HTTP request.
    1. Wait for the channel to close.
    1. Hide the spinner.
 
This second option means that HTTP requests longer than your sleep don't take any longer than they would have without the sleep. Really fast requests will still show the spinner for the sleep time.

jstangroome: Frankly I feel that CLIs that move the cursor backward to manipulate previously written output are unnecessarily complex. But this is totally my personal take.

auxesis: Here's a prototype of option 1 (transitioning to a completion message): 

new spinner

WDYT @jstangroome?

jstangroome: If it means we no longer make requests take longer arbitrarily 👍

PR #61, opened 2020-12-17
Allow variable Node.js directory paths, default to 'nodejs'

interludic2000:
Allow `sectionctl deploy` and `sectionctl ps` to specify the folder path of NodeJS module in Section repo. 
This is to allow users to use something other than the default path of `nodejs`

### DIFF
--- a/api/applications.go
+++ b/api/applications.go
@@ -228,7 +228,7 @@ func getEnvironmentID(accountID int, applicationID int, environmentName string) 
 }
 
 // ApplicationStatus returns a module's current status on Section's delivery platform
-func ApplicationStatus(accountID int, applicationID int) (as []AppStatus, err error) {
+func ApplicationStatus(accountID int, applicationID int, moduleName string) (as []AppStatus, err error) {
 	u := BaseURL()
 	u.Path = "/new/authorized/graphql_api/query"
 
@@ -246,7 +246,7 @@ func ApplicationStatus(accountID int, applicationID int) (as []AppStatus, err er
 	}
 
 	requestData.Variables = map[string]interface{}{
-		"moduleName":    "nodejs",
+		"moduleName":    moduleName,
 		"environmentID": environmentID,
 	}
 	requestData.Query = "query DeploymentStatus($moduleName: String!, $environmentID: Int!){deploymentStatus(moduleName:$moduleName, environmentID:$environmentID){inService state instanceName payloadID}}"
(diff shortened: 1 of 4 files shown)


api/applications.go
```diff
@@ -228,7 +228,7 @@ func getEnvironmentID(accountID int, applicationID int, environmentName string)
 }
 
 // ApplicationStatus returns a module's current status on Section's delivery platform
-func ApplicationStatus(accountID int, applicationID int) (as []AppStatus, err error) {
+func ApplicationStatus(accountID int, applicationID int, moduleName string) (as []AppStatus, err error) {
```
pavelnikolov, line 231: I guess that you need a `ctx context.Context` as a first argument. This can't run forever so a reasonable timeout would be a good fit.

pavelnikolov: Maybe outside of the scope of this PR.

mwickman:
Is this all that's needed to complete the story so non nodejs named repos can use sectionctl? I had the impression there would be server-side work. Should we also make a task to update docs?

auxesis:
@mwickman yep, @interludic2000 worked out that this is literally the only change required to support this use case. 🎉

mwickman (approved):
Tested and worked like a charm for me, I created a new node app called `cartjs` just fine and watched it deploy with the `ps` command. Nice work! 
image

commands/deploy.go
```diff
@@ -160,7 +161,7 @@ func (c *DeployCmd) Run() (err error) {
 			},
 		},
 	}
-	err = api.ApplicationEnvironmentModuleUpdate(c.AccountID, c.AppID, c.Environment, "nodejs/.section-external-source.json", ups)
+	err = api.ApplicationEnvironmentModuleUpdate(c.AccountID, c.AppID, c.Environment, c.AppPath + "/.section-external-source.json", ups)
```
jstangroome, line 164: [path.Join()](https://golang.org/pkg/path/#Join) would be better than string concatenation here for the case where the user has included the trailing slash in their argument value.

commands/deploy.go
```diff
@@ -35,6 +35,7 @@ type DeployCmd struct {
 	ServerURL   *url.URL      `default:"https://aperture.section.io/new/code_upload/v1/upload" help:"URL to upload application to"`
 	Timeout     time.Duration `default:"300s" help:"Timeout of individual HTTP requests."`
 	SkipDelete  bool          `help:"Skip delete of temporary tarball created to upload app."`
+	AppPath     string        `default:"nodejs" help:"Path of NodeJS application in environment repository."`
```
jstangroome, line 38: I imagine the server API will protect against invalid characters in this argument but:

1. is the server error about why the apppath is wrong communication back to the user in a helpful form? and/or
1. should we validate the apppath locally and tell the user quickly instead of waiting until after server communication completes, especially the upload?

PR #62, opened 2020-12-17
Make tokens work when no stored creds

auxesis:
A pretty major internal refactor, but: 

This fixes a bug where a token provided in the `SECTION_TOKEN` environment variable won't be used if there is also no token stored in the keyring. 

Currently, this bug makes `SECTION_TOKEN` is near unusable in CI environments, or environments without a keyring (like Linux without DBUS).


credentials/credentials.go
```diff
@@ -18,24 +18,23 @@ var (
 )
 
 // Setup ensures authentication is set up
-func Setup(endpoint string) (err error) {
+func Setup(endpoint string) (token string, err error) {
```
pavelnikolov, line 19: why do you use named return values? No need for that.

credentials/credentials.go
```diff
@@ -18,24 +18,23 @@ var (
...
 	if !IsCredentialRecorded(KeyringService, KeyringUser) {
 		fmt.Printf("No API credentials recorded.\n\n")
 		fmt.Printf("Let's get you authenticated to the Section API!\n\n")
 
-		t, err := PromptForCredential(os.Stdin, os.Stdout)
+		t, err := Prompt(os.Stdin, os.Stdout)
 		if err != nil {
-			return fmt.Errorf("error when prompting for credential: %w", err)
+			return token, fmt.Errorf("error when prompting for credential: %w", err)
```
pavelnikolov: This makes me look where `token` is assigned a value.

credentials/credentials.go
pavelnikolov: Unless you set the value in a deferred statement after the `return`

mwickman (approved):
lgtm but it appears the tests aren't passing from the CI check

mwickman:
Here are screenshots of Nadiv's WSL vs my Linux system. I run dbus and he does not. It appears to me that without some sort of keyring software running, we get these storage errors. It doesn't appear simple to securely store a token without a keyring available, so I'm unsure if we should make a workaround, though we'd at least need to fix the build. We could also consider making a more instructive error message for the user.

image (48)
image (47)

credentials/credentials.go
auxesis: It's a stylistic choice. 

I find it makes it clearer what variables are going to be returned.

auxesis:
@mwickman yep, so if you set the `SECTION_TOKEN` environment variable, `sectionctl` will skip looking up a credential in the keyring. 

I noticed this behaviour yesterday when debugging with @interludic2000, and that's what this bugfix PR is for.

auxesis:
There's a way to put the keyring into a mocking mode during tests, but the tests aren't doing it here. Need to fix that.

credentials/credentials.go
jstangroome: I gotta subscribe to the Dave Cheney ["Avoid named return values" rule](https://dave.cheney.net/practical-go/presentations/gophercon-israel.html#_avoid_named_return_values) for practical Go.

Also [A Tour of Go](https://tour.golang.org/basics/7):
>Naked return statements should be used only in short functions, as with the example shown here. They can harm readability in longer functions.

This probably meets the "short function" criteria to be acceptable but care should be taken... or make or your functions much shorter ;)

PR #63, opened 2020-12-22
Raise auth specific errors

auxesis:
Add explicit errors for authentication failures from the API. 

Makes it clearer for API client users if requests are due to auth failures.


mwickman (commented):
LGTM

PR #67, opened 2021-01-12
Support supplying a token on the command line when logging in

auxesis:
Works like this:

```
read -es SECTION_TOKEN
sectionctl login
```

Or less securely (because the secret is stored in shell history): 

``` bash
sectionctl login --section-token s3cr3t:t0k3n
```

### DIFF
--- a/commands/login.go
+++ b/commands/login.go
@@ -18,11 +18,18 @@ type LoginCmd struct {
 
 // Run executes the command
 func (c *LoginCmd) Run() (err error) {
-	t, err := credentials.PromptAndWrite(c.In(), c.Out(), api.PrefixURI.Host)
-	if err != nil {
-		return fmt.Errorf("unable to prompt and write credentials: %w", err)
+	if api.Token != "" {
+		err = credentials.Write(api.PrefixURI.Host, api.Token)
+		if err != nil {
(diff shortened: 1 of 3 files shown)


jstangroome (approved):

jstangroome:
Ideally we'd encourage token specification methods that don't persist in shell history, or expose them to casual observation of the process tree.

Spitballing: Perhaps a `--token-from-stdin` option? E.g.:
```sh
read -es TOKEN
printf %s "$TOKEN" | sectionctl login --token-from-stdin
```
I'm sure there are cleaner methods and suitable PowerShell alternatives.

You'll note that all the cloud container registry integrations with `docker` stopped passing creds via CLI args and switched to more secure methods.

auxesis:
Fun fact: this PR already indirectly supports that, by setting the `SECTION_TOKEN` environment variable:

```
read -es SECTION_TOKEN
sectionctl login
```

jstangroome:
> Fun fact: this PR already indirectly supports that, by setting the `SECTION_TOKEN` environment variable

IIRC, environment variables are also casually observable via `cat /proc/$(pidof sectionctl)/environ` but I'm not going to die on this particular hill. 😄  🚶

PR #68, opened 2021-01-12
Add app create and delete subcommands

auxesis:
Adds two new `apps` subcommands:

```
  create    Create new app on Section.

  delete    Delete an existing app on Section.
```


mwickman (approved):
Some unit tests would be ideal but otherwise LGTM

auxesis:
@mwickman have added tests for failure modes, so good to go.

PR #69, opened 2021-01-13
Don't require user to be logged in to run version

auxesis:
Before: 

If a user runs `sectionctl version` and they haven't logged in with `sectionctl login`, they will be presented a login prompt before they see the version string.

After: 

The version string is emitted immediately, and auth isn't checked.

### DIFF
--- a/sectionctl.go
+++ b/sectionctl.go
@@ -48,6 +48,7 @@ func bootstrap(c CLI, ctx *kong.Context) {
 	log.SetOutput(filter)
 
 	switch {
+	case ctx.Command() == "version":
 	case ctx.Command() == "login":
 		api.Token = c.SectionToken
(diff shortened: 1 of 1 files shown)


sectionctl.go
```diff
@@ -48,6 +48,7 @@ func bootstrap(c CLI, ctx *kong.Context) {
 	log.SetOutput(filter)
 
 	switch {
+	case ctx.Command() == "version":
```
jstangroome, line 51: Worth being more explicit about the empty case?
```suggestion
	case ctx.Command() == "version":
		// bypass auth check for version command
```

jstangroome (approved):

sectionctl.go
auxesis: That's a great call, I'll add that now.

PR #73, opened 2021-01-14
Increase and make timeouts configurable

auxesis:
Increase the timeout on patch requests to update module configuration to 5 minutes, as server side validation of changes can take a while. 

Make request timeouts globally configurable with `--section-api-timeout`.

### DIFF
--- a/api/applications.go
+++ b/api/applications.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"log"
 	"net/http"
+	"time"
 )
 
 // App represents an application deployed on Section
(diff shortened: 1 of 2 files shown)


api/applications.go
```diff
@@ -228,6 +229,7 @@ func ApplicationEnvironmentModuleUpdate(accountID int, applicationID int, env st
 	}
 	log.Printf("[DEBUG] JSON payload: %s\n", b)
 	headers := map[string][]string{"filepath": []string{filePath}}
+	Timeout = 300 * time.Second // five minutes, because server-side validation can take a while
```
pavelnikolov, line 232: Where is this `Timeout` being used?

api/applications.go
```diff
@@ -228,6 +229,7 @@ func ApplicationEnvironmentModuleUpdate(accountID int, applicationID int, env st
...
 	resp, err := request(http.MethodPatch, u, bytes.NewBuffer(b), headers)
```
pavelnikolov, line 233: Shouldn't `request` accept `context.Context` with timeout?

auxesis: Yeah, you're right, my implementation here is crappy, and I should be using `context.Context`.

I'll refactor this now.

auxesis:
I'm going to merge this PR now, and put up a separate PR that refactors timeouts to use `context.Context`.

PR #76, opened 2021-01-18
Introduces follow command when viewing logs

manibatra:
- Sleeps for 5 seconds between every run
- Timestamp is iterated by one second if latest timestamp is not found in response.

https://user-images.githubusercontent.com/2607129/104877103-35b1ef80-59ad-11eb-9945-bbafea8edea3.mov


commands/logs.go
```diff
@@ -15,46 +16,74 @@ const maxNumberLogs = 1500
...
+			} else if a.Type == "access" {
+				log.Printf("%s%s\t%s\n", aurora.Green(a.InstanceName), aurora.Green("["+a.Type+"]"), a.Message)
+			} else {
+				log.Printf("%s[%s]\t%s\n", a.InstanceName, a.Type, a.Message)
+			}
+		}
+		if !c.Follow {
+			break
+		}
+		latestTimestamp := getLatestTimeStampFromLogs(appLogs)
```
jstangroome: Instead of calling `getLatestTimeStampFromLogs()` and iterating the array twice, you could just track the timestamp as you emit the logs:
```suggestion
			if a.Timestamp != "" {
				latestTimestamp = a.Timestamp
			}
		}
		if !c.Follow {
			break
		}
```

commands/logs.go
```diff
@@ -15,46 +16,74 @@ const maxNumberLogs = 1500
...
+		if latestTimestamp == "" {
+			latestTimestamp = startTimestampRfc3339
+		}
+		t, err := time.Parse(time.RFC3339, latestTimestamp)
+		if err == nil {
+			t = t.Add(time.Second)
```
jstangroome, line 83: I'm not sure that continuing to increment the start timestamp like this is logical and if anything will lead to dropped logs.

commands/logs.go
```diff
@@ -15,46 +16,74 @@ const maxNumberLogs = 1500
...
+			startTimestampRfc3339 = t.Format(time.RFC3339)
 		}
+		s.Prefix = ""
+		s.FinalMSG = ""
+		s.Start()
+		time.Sleep(5 * time.Second)
```
jstangroome, line 89: Why five seconds?

commands/logs.go
```diff
@@ -15,46 +16,74 @@ const maxNumberLogs = 1500
...
+	Length       int    `default:100 help:"Number of log lines to fetch."`
+	Follow       bool   `help:"Displays recent logs and leaves the session open for logs to stream in. --instance-name required."`
 	// StartTimestamp  int    `default:0 help:"Start of log time stamp to fetch."`
 	// EndTimestamp    int    `default:0 help:"End of log time stamp to fetch."`
 }
 
 // Run executes the command
 func (c *LogsCmd) Run() (err error) {
 	s := NewSpinner("Getting logs from app")
+	logsHeader := "\nInstanceName[Log Type]\t\t\tLog Message\n"
```
jstangroome, line 32: Probably for a separate PR but it seems to me that if the user specified an exact instance name in the args, its redundant to repeat that instance name back to the user on every single log line emitted. 🤷

jstangroome (approved):

commands/logs.go
```diff
@@ -15,46 +16,74 @@ const maxNumberLogs = 1500
...
+		if err != nil {
+			return err
+		}
+		for _, a := range appLogs {
+			a.Message = strings.TrimSpace(a.Message)
+
+			if a.Type == "app" {
+				log.Printf("%s%s\t%s\n", aurora.Cyan(a.InstanceName), aurora.Cyan("["+a.Type+"]"), a.Message)
+			} else if a.Type == "access" {
+				log.Printf("%s%s\t%s\n", aurora.Green(a.InstanceName), aurora.Green("["+a.Type+"]"), a.Message)
```
jstangroome, line 67: Is it in the backlog for sectionctl to json-decode access logs and render them human readable?

commands/logs.go
manibatra: Is to prevent duplicate logs in case there are no logs returned while polling. There are a few edge cases that we are aware might happen. But wanted to get this in the hands of the users first.

commands/logs.go
manibatra: Good call. I will fix this and remove the code not required.

commands/logs.go
manibatra: Just arbitrary at the moment till we see the real world usage. I will add a comment.

commands/logs.go
manibatra: I believe the intent is to keep the formatting consistent with when the instance name is not used. Heroku being the inspiration.

commands/logs.go
manibatra: I will raise it in the next meeting. Not that I am aware of.

commands/logs.go
jstangroome: I'd be more inclined to store the last element of `appLogs` and if the same log appears in the next api response, only emit logs after that entry.

PR #77, opened 2021-01-19
Add friendlier messaging for reaching rate limit

interludic2000:
When a status 429 is received from Section API, display a friendly error message.

image

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -76,6 +76,10 @@ func request(method string, u url.URL, body io.Reader, headers ...http.Header) (
 
 // prettyTxIDError creates a support friendly error message with an transaction ID
 func prettyTxIDError(resp *http.Response) error {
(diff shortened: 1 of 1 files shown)


manibatra (approved):

api/api.go
```diff
@@ -76,6 +76,10 @@ func request(method string, u url.URL, body io.Reader, headers ...http.Header) (
 
 // prettyTxIDError creates a support friendly error message with an transaction ID
 func prettyTxIDError(resp *http.Response) error {
+	if resp.Status == "429 Too Many Requests"  {
```
jstangroome, line 79: The preferred, compiler-checked code would be:
```!suggestion
	if resp.StatusCode == http.StatusTooManyRequests {
```
https://golang.org/pkg/net/http/#pkg-constants

PR #79, opened 2021-01-19
Fix length shorthand and rate limit string match

interludic2000:
* Fixed number parameter shorthand which was broken due to previous PR.
* Implement better string matching for status 429.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"net/url"
 	"runtime"
+	"strconv"
 	"time"
 
 	"github.com/section/sectionctl/version"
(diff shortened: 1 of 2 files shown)


manibatra (approved):

api/api.go
```diff
@@ -76,7 +77,7 @@ func request(method string, u url.URL, body io.Reader, headers ...http.Header) (
 
 // prettyTxIDError creates a support friendly error message with an transaction ID
 func prettyTxIDError(resp *http.Response) error {
-	if resp.Status == "429 Too Many Requests"  {
+	if resp.Status == strconv.Itoa(http.StatusTooManyRequests) + " " + http.StatusText(http.StatusTooManyRequests)  {
```
jstangroome, line 80: Why does the string need to match? Why not just `resp.StatusCode` integer?? HTTP servers are free to put any arbitrary text after the integer. A `404 Not Found` response line and a `404 Where's it at?` response line is equivalent AFAIK.

PR #82, opened 2021-01-19
Make request timeouts explicit with context

auxesis:
As promised in #73, this PR makes timeouts explicit by setting them on a `context.Context`.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -41,7 +41,7 @@ func BaseURL() (u url.URL) {
 //
 // You can pass 0 or more headers, and keys in the later headers will override earlier passed headers.
 func request(ctx context.Context, method string, u url.URL, body io.Reader, headers ...http.Header) (resp *http.Response, err error) {
-	var client http.Client{}
+	var client http.Client
 
 	req, err := http.NewRequestWithContext(ctx, method, u.String(), body)
 	if err != nil {
--- a/api/api.go
+++ b/api/api.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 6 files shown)


api/api.go
```diff
@@ -39,12 +40,10 @@ func BaseURL() (u url.URL) {
 // request does the heavy lifting of making requests to the Section API.
 //
 // You can pass 0 or more headers, and keys in the later headers will override earlier passed headers.
-func request(method string, u url.URL, body io.Reader, headers ...http.Header) (resp *http.Response, err error) {
-	client := &http.Client{
-		Timeout: Timeout,
-	}
+func request(ctx context.Context, method string, u url.URL, body io.Reader, headers ...http.Header) (resp *http.Response, err error) {
+	client := http.Client{}
```
pavelnikolov: (nitpick) Default value works fine.
```go
var client http.Client
```

pavelnikolov (approved):

mwickman (approved):

api/api.go
pavelnikolov: Also, why don't you reuse the client across different quests?

auxesis: Yep that's a good point about the default value, I don't know why I didn't do it.

auxesis: @pavelnikolov I wasn't re-using the client because I hadn't thought of it! 

I hadn't thought of it because I was setting up a client with some default values (like timouts) on every call to `request()`

Now we're using the default value, and timeouts are handled inside `context`s, 14fbeac96f8b92b24c73ea6121a9d957c6cff5bf moves `client` to a package level variable, so it can be reused.

PR #83, opened 2021-01-19
Remove debug output from logs output

manibatra:


auxesis (approved):
LGTM (but haven't tested on Windows).

PR #85, opened 2021-01-28
Add version checking

auxesis:
Works like this: 

https://user-images.githubusercontent.com/12306/106137398-c2835700-61be-11eb-9637-a1e1e508415e.mov


AskAlice (approved):

commands/version.go
```diff
@@ -1,17 +1,124 @@
...
+		return
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		errs <- fmt.Errorf("bad response from GitHub. Please try again later")
+		return
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
```
pavelnikolov, line 93: you can use directly a `json.Decoder`:

```go
err := json.NewDecoder(resp.Body).Decode(&latestResp)
if err != nil {
    return err
}
// happy path

pavelnikolov (approved):

PR #87, opened 2021-01-29
Add domains subcommand

auxesis:
For listing domains on an account

### DIFF
--- a/api/domains.go
+++ b/api/domains.go
@@ -9,6 +9,50 @@ import (
 	"time"
 )
 
+// DomainsResponse represents an API response to GET /account/{accountId}/domains
+type DomainsResponse struct {
+	DomainName string `json:"domain_name"`
+	Engaged    bool   `json:"engaged"`
+}
+
+// Domains returns a list of an account's domains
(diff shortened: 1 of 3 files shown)


api/domains.go
```diff
@@ -9,6 +9,50 @@ import (
...
+		case 401:
+			return d, ErrStatusUnauthorized
+		case 403:
+			return d, ErrStatusForbidden
+		default:
+			return d, prettyTxIDError(resp)
+		}
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
```
pavelnikolov, line 43: `ioutil` is getting deprecated, no?

pavelnikolov (approved):

api/domains.go
auxesis: Yep it is 🎉 

Right now I'm aiming for consistency with the rest of the code base. 

I'll put up a PR to upgrade all the legacy packages once 1.16 drops.

pavelnikolov: In this case you don't really need it:

```go
err := json.NewDecoder(resp.Body).Decode(&d)
return d, err
```

PR #89, opened 2021-02-03
Adds sectionctl init command to initialize new apps with the required files for deployment.  

nadivgold:
### Description

`sectionctl init` takes one optional argument `stack-name` that defaults to `nodejs-basic`. The command creates the `server.conf` file, or validates the existing one. The command also creates a `package.json` if it does not exist, of if there is an existing `package.json`, provides a warnings about start script fields.


commands/init.go
```diff
@@ -0,0 +1,102 @@
...
+		} else {
+			fmt.Println("Validating package.json")
+			fileinfo, err := checkPkgJSON.Stat()
+			if err != nil {
+				panic(err)
+			}
+			buff := make([]byte, fileinfo.Size())
+			checkPkgJSON.Read(buff)
+			fileString := string(buff)
+			if !strings.Contains(fileString, "name") {
```
AskAlice: name and version are no longer required, but start is still required yeah.

commands/init.go
```diff
@@ -0,0 +1,102 @@
...
+	proxy_set_header X-Forwarded-Proto $http_x_forwarded_proto;
+	proxy_set_header Host $host;
+	include /etc/nginx/section.module/node.conf;
+}
+		
+location ~ "/notnuxt/" {
+	proxy_set_header X-Forwarded-For $http_x_forwarded_for;
+	proxy_set_header X-Forwarded-Proto $http_x_forwarded_proto;
+	proxy_set_header Host $host;
+	proxy_pass http://next-hop;
```
AskAlice: I don't think this is necessary or useful unless they are using proxies that are higher up in the chain. Most people using sectionctl probably aren't initializing their repo with such an advanced config, but the next hop capability should definitely be documented in our docs.

AskAlice (commented):
Looks good to me, though I don't think I'm the one to approve/merge this.

commands/init.go
nadivgold: Okay good to know. I can remove that from the PR

commands/init.go
nadivgold: I've included it because it's in the `server.conf` in the demo app. I think it's safe to keep in in the event that they want additional proxies in the future. Is there a reason to remove it?

AskAlice: at the very least it shouldn't be called /notnuxt/ for a general application of it

nadivgold: So I should remove this entire location part?

```
location ~ "/notnuxt/" {
	proxy_set_header X-Forwarded-For $http_x_forwarded_for;
	proxy_set_header X-Forwarded-Proto $http_x_forwarded_proto;
	proxy_set_header Host $host;
	proxy_pass http://next-hop;
    }`)
}
```

AskAlice: Maybe comment it out to show it's not required, and just be like `/next-proxy-hop/`

commands/apps.go
```diff
@@ -169,3 +173,121 @@ func (c *AppsDeleteCmd) Run() (err error) {
 
 	return err
 }
+
+// AppsInitCmd creates and validates server.conf and package.json to prepare an app for deployment
+type AppsInitCmd struct {
+	StackName string `optional default:"nodejs-basic" short:"s" help:"Name of stack to deploy. Default is nodejs-basic"`
+	Force     string `optional default:"false" short:"f" help:"Resets deployment specific files to their default configuration"`
```
auxesis: This can be a boolean.

commands/apps.go
```diff
@@ -169,3 +174,130 @@ func (c *AppsDeleteCmd) Run() (err error) {
...
+	proxy_pass http://next-hop;
+}`)
+}
+
+// Run executes the command
+func (c *AppsInitCmd) Run() (err error) {
+	var stdout bytes.Buffer
+	var stderr bytes.Buffer
+	switch c.StackName {
+	case "nodejs-basic":
```
AskAlice, line 206: is this necessary for now?

I imagine we might have other stacks if we ever do like... static site hosting or python/django as a module?

Also, maybe just like `-s node` or `--stack node` would be easier to write

nadivgold: it's for future proofing, for when we offer more than node or hosting varnish. end users won't have to worry about this for now since the command defaults to 'nodejs-basic'

PR #92, opened 2021-02-11
Fix ignore directories and filemode of node binaries on windows

AskAlice:
This will fix the way it ignores directories, before it was assuming there was always a forwardslash, but windows uses backslashes.

Also it will ensure anything in the node_modules/.bin/ folder is executable, as windows doesn't set filemode.

### DIFF
--- a/commands/deploy.go
+++ b/commands/deploy.go
@@ -279,7 +279,6 @@ func addFileToTarWriter(filePath string, tarWriter *tar.Writer, prefix string) e
 		match := strings.Contains(baseFilePath,"node_modules\\.bin")
 		if match {
 			header.Mode = 511 // This is the value that it spat out when i ran npm i on linux then checked the int64 value of the stat.Mode() on files in node_modules/.bin/
-			fmt.Printf("added %s\n", baseFilePath);
 		}
 	}
 	err = tarWriter.WriteHeader(header)
--- a/commands/deploy.go
+++ b/commands/deploy.go
@@ -273,7 +273,14 @@ func addFileToTarWriter(filePath string, tarWriter *tar.Writer, prefix string) e
 	// must provide real name
 	// (see https://golang.org/src/archive/tar/common.go?#L626)
 	header.Name = filepath.ToSlash(baseFilePath)
-
+  // ensure windows provides filemodes for binaries in node_modules/.bin
+	if runtime.GOOS == "windows" {
+		match := strings.Contains(baseFilePath,"node_modules\\.bin")
+		if match == true {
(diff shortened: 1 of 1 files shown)


commands/deploy.go
```diff
@@ -194,7 +195,10 @@ func IsValidNodeApp(dir string) (errs []error) {
 
 	return errs
 }
-
+// Split helps differentiate between different directory delimiters. / or \
+func Split(r rune) bool {
```
jstangroome, line 199: Why not [filepath.Split()](https://golang.org/pkg/path/filepath/#Split)?
>for manipulating filename paths in a way compatible with the target operating system-defined file paths.

commands/deploy.go
```diff
@@ -268,7 +274,14 @@ func addFileToTarWriter(filePath string, tarWriter *tar.Writer, prefix string) e
 	// must provide real name
 	// (see https://golang.org/src/archive/tar/common.go?#L626)
 	header.Name = filepath.ToSlash(baseFilePath)
-
+  // ensure windows provides filemodes for binaries in node_modules/.bin
+	if runtime.GOOS == "windows" {
+		match := strings.Contains(baseFilePath,"node_modules\\.bin")
+		if match == true {
+			header.Mode = 511 // This is the value that it spat out when i ran npm i on linux then checked the int64 value of the stat.Mode() on files in node_modules/.bin/
```
jstangroome: Why not octal literals? 511 decimal would be `0o777`. That said, `0o755` (493 decimal) would be a better mode to use.
https://golang.org/ref/spec#Integer_literals

commands/deploy.go
AskAlice: TIL that exists. I just used google.

commands/deploy.go
AskAlice: I set it to what npm i set the filemode to be.

I agree 493 would be better, but i'm not quite sure who the owner is when the tarball is extracted.

AskAlice: @jstangroome how did you know that this int64 value is an octal literal?
I just reverse engineered the expected value via strconv/printf

jstangroome: File mode is traditionally octal. 🤷

commands/deploy.go
AskAlice: It doesn't split the dir into an array.

Take for example if someone has a monorepo with multiple .lint folders, or if we later add the ability to specify ignored directories as a flag

auxesis (approved):
Thanks for these fixes @AskAlice.

PR #98, opened 2021-02-15
Deploy via git

AskAlice:
The git hook pre-receive doesn't show when you use a json patch, which often causes an EOF while waiting for a response from the API.

This instead uses git to show the output of the pre-receive, which can be used to provide debug information to the user rather than a useless/vague error that has nothing to do with what actually went wrong.


AskAlice:
I'm not quite sure how I should approach making deploy_test.go work with this setup, but it works to deploy this way. 

Furthermore I wasn't really tasked with this, I just felt like I really needed it to do some of my assigned tasks efficiently and wrote this sunday night. So I don't feel like I should write a deploy_test anytime soon, and I doubt that we should merge immediately.

However it is quite helpful as it will tell you what went wrong in the git hook pre-receive rather than just returning an http error.

http://emu.sh/upload.webm

commands/deploy.go
```diff
@@ -150,27 +152,117 @@ func (c *DeployCmd) Run() (err error) {
...
+	err = updateGitViaGit(c, response);
 	if err != nil {
 		return fmt.Errorf("failed to trigger app update: %v", err)
 	}
 
+	// err = api.ApplicationEnvironmentModuleUpdate(c.AccountID, c.AppID, c.Environment, c.AppPath+"/.section-external-source.json", ups)
+	// if err != nil {
+	// 	return fmt.Errorf("failed to trigger app update: %v", err)
+	// }
+
```
jstangroome, line 165: Just delete code, git history will keep it.

mwickman:
Alice and I discussed on a call and we decided to park this PR for the time being. In my opinion, we should fix the PATCH request to behave as we want instead of adding new git-like behaviors to sectionctl. Leaving the branch up as it can be useful for local development in the meantime.

PR #99, opened 2021-02-16
Write debug logs to file

auxesis:
When you run `sectionctl` with the `--debug` flag set, also write the debug output to a file. 

This makes it easier for users to report information about their environment when sending support requests.

### DIFF
--- a/sectionctl.go
+++ b/sectionctl.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"fmt"
+	"io"
 	"log"
 	"net/url"
 	"os"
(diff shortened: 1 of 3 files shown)


sectionctl.go
```diff
@@ -49,6 +51,18 @@ func bootstrap(c CLI, ctx *kong.Context) {
...
+		logFile, err := os.OpenFile(logFilename, os.O_CREATE|os.O_APPEND|os.O_RDWR, 0666)
+		if err != nil {
+			panic(err)
+		}
+		fmt.Fprintf(logFile, "Version:   %s\n", commands.VersionCmd{}.String())
+		fmt.Fprintf(logFile, "Command:   %s\n", ctx.Args)
+		fmt.Fprintf(logFile, "PrefixURI: %s\n", api.PrefixURI)
+		fmt.Fprintf(logFile, "Timeout:   %s\n", api.Timeout)
+		fmt.Printf("Writing debug log to: %s\n", logFilename)
+		mw := io.MultiWriter(logFile, colorableWriter)
```
AskAlice, line 64: so this outputs it to the PWD? 

Perhaps it could go to `~/sectionctl-debug-%s-log` and let the user know?

AskAlice (approved):

sectionctl.go
auxesis: The least surprising behavior will be to write it in the current working directory. 

We already write out the path to the debug filename as soon as you run `sectionctl` with the `--debug` flag. 

If the user misses the filename in the output, they'll at least have a file in their current working directory.

PR #101, opened 2021-02-16
Make sectionctl apps init more robust

nadivgold:
sectionctl apps init now uses JSON parsing to manipulate package.json.  

Errors are now handled by the parent function. 

There are testing tables that check for various different potential errors.


commands/apps.go
```diff
@@ -293,49 +305,75 @@ func (c *AppsInitCmd) InitializeNodeBasicApp(stdout, stderr bytes.Buffer) (err e
...
+		return fmt.Errorf("package.json is not valid JSON %w", err)
 	}
-	fStr := string(buf)
-	if !strings.Contains(fStr, `"start": "`) {
-		replace := `"scripts": {
-    "start": "node YOUR_SERVER_HERE.js",`
-		fStr = strings.Replace(fStr, `"scripts": {`, replace, 1)
-		err = validPkgJSON.Truncate(0)
+	log.Println(jsonMap["scripts"])
+	lv := jsonMap["scripts"]
```
AskAlice, line 346: What if scripts is empty? This will error.

commands/apps.go
```diff
@@ -293,49 +305,75 @@ func (c *AppsInitCmd) InitializeNodeBasicApp(stdout, stderr bytes.Buffer) (err e
...
+	err = json.Unmarshal(buf, &jsonMap)
 	if err != nil {
-		panic(err)
+		log.Println("[ERROR] JSON format invalid for package.json")
+		return fmt.Errorf("package.json is not valid JSON %w", err)
 	}
-	fStr := string(buf)
-	if !strings.Contains(fStr, `"start": "`) {
-		replace := `"scripts": {
-    "start": "node YOUR_SERVER_HERE.js",`
```
AskAlice, line 325: why not use an interface and json.MarshalIndent?

commands/apps.go
```diff
@@ -293,49 +305,75 @@ func (c *AppsInitCmd) InitializeNodeBasicApp(stdout, stderr bytes.Buffer) (err e
...
+	lv := jsonMap["scripts"]
+	stringMap, ok := lv.(map[string]interface{})
+	if !ok {
+		return fmt.Errorf("json unable to be read as map[string]interface %w", err)
+	}
+	_, ok = stringMap["start"]
+	if !ok {
+		stringMap["start"] = "node YOUR_SERVER_HERE.js"
+		jsonMap["scripts"] = stringMap
+		err = os.Truncate("package.json", 0)
```
AskAlice, line 355: If there is no start script, it makes the package.json empty? That will break so many users repos if it errors.

Also, have you tested this on the package.jsons of the frameworks on the deploy page?

https://aperture.section.io/new/configure/setup/account/0/application/0/environment/0/frameworks

AskAlice:
In summation I don't think we should be destructive to the package.json at all. I think it would be better to tell the user that the start script is required to run on production, so they should add one themselves.

AskAlice:
When i ran this on an app that lacked a start script but had everything else, I got this output, see line 1:
image

commands/apps.go
```diff
@@ -293,49 +305,75 @@ func (c *AppsInitCmd) InitializeNodeBasicApp(stdout, stderr bytes.Buffer) (err e
...
+	buf, err := ioutil.ReadFile("package.json")
 	if err != nil {
-		panic(err)
+		return fmt.Errorf("failed to read package.json %w", err)
 	}
-	buf := make([]byte, finfo.Size())
-	_, err = validPkgJSON.Read(buf)
+	fStr := string(buf)
+	if fStr == "" {
+		err := os.Remove("package.json")
```
AskAlice, line 325: this feels dirty. Maybe check if the length of fstr is zero, or better, get the filesize, then yes/no prompt the user

commands/apps.go
nadivgold: Oops that was testing code that snuck its way in

commands/apps.go
nadivgold: That's what this PR does. I used to have it do string manipulation but that was before I learned about the json parser library and map[string]interface

commands/apps.go
nadivgold: I read in the entire file into memory, modify it in memory, then write it to the disk. I empty the package.json file in the disk and then write in the updated code. Nothing should be lost. the only thing that should change is the addition of a start script with a template for what to put into it.

nadivgold:
> When i ran this on an app that lacked a start script but had everything else, I got this output, see line 1:
> image

I thought I fixed that. I'll take another look. 

edit: can not reproduce this

commands/apps.go
nadivgold: I can use the length. Is that functionally different from checking to see if it's the empty string?

nadivgold:
> In summation I don't think we shouldn't be destructive to the package.json at all. I think it would be better to tell the user that the start script is required to run on production, so they should add one themselves.

That is basically what this does with the additional step of adding in a template that looks like:
`"start": "node YOUR_SERVER_HERE.js"`

commands/apps_test.go
```diff
@@ -45,3 +47,137 @@ func TestCommandsAppsCreateAttemptsToValidateStackOnError(t *testing.T) {
...
+	}
+	defer f.Close()
+	_, err = f.WriteString(data)
+	if err != nil {
+		return err
+	}
+	return err
+}
+
+func TestInitNodejsBasicAppEncounteringPossibleFailureStates(t *testing.T) {
```
auxesis: Great to see some tests! 

The test name should: 

- namespaced (`TestCommandsAppsInit`, so you can run a subset of tests with `go test ./... -run TestCommandsAppsInit`), and
- more direct

How about `TestCommandsAppsInitHandlesErrors`?

commands/apps_test.go
```diff
@@ -45,3 +47,137 @@ func TestCommandsAppsCreateAttemptsToValidateStackOnError(t *testing.T) {
...
+	assert := assert.New(t)
+
+	// Setup
+	var testCases = []struct {
+		servConf string
+		pkgJson  string
+		isFatal  bool
+	}{
+		{`location / {
```
auxesis: Let's use `helperLoadBytes` to clean up these tests. 

That way we can put the test data in `commands/testdata/apps/init/` and move data out of these tests.

commands/apps_test.go
```diff
@@ -45,3 +47,137 @@ func TestCommandsAppsCreateAttemptsToValidateStackOnError(t *testing.T) {
...
+
+	cmd := AppsInitCmd{
+		StackName: "nodejs-basic",
+		Force:     false,
+	}
+	for _, tc := range testCases {
+		n := ""
+		t.Run(n, func(t *testing.T) {
+			// Setup
+			err := OverwriteFile("server.conf", tc.servConf)
```
auxesis: There are a few big drawbacks to this approach:

- If a test in the table fails, the data fed into the test will be blown away by subsequent tests in the table that are run (because we're reading and writing the same file in each test case)
- If I'm writing a test and I decide to overwrite `sectionctl.go` or `api/api.go`, I'm gonna have a bad time. 

The better way to handle this is to create a temporary directory for the test case to run in. 

There's a [good example of this](https://github.com/section/sectionctl/blob/583716bfef5195e3d7ae3b54031cf2a7be81c89f/commands/deploy_test.go#L109-L111) in the `deploy` subcommand tests.

commands/apps_test.go
auxesis: Here's [an example of this approach](https://github.com/section/sectionctl/blob/583716bfef5195e3d7ae3b54031cf2a7be81c89f/commands/apps_test.go#L27) inside this test file itself.

commands/apps.go
AskAlice: ah you're right, don't know why i was reading red.

commands/apps_test.go
auxesis: Will clean up the test cases to look like this:

``` golang
var testCases = []struct {
	servConf string
	pkgJson  string
	isFatal  bool
}{
	{ string(helperLoadBytes(t, "apps/init.good.server.conf`")),  string(helperLoadBytes(t, "apps/init.bad.package.json`")), false },
	{ string(helperLoadBytes(t, "apps/init.bad.server.conf`")),  string(helperLoadBytes(t, "apps/init.bad.package.json`")), true },
}
```

auxesis (changes requested):
Thanks for putting this PR together @nadivgold. 

I've got a bunch of feedback about how the tests are structured that needs to be addressed before we can merge this. 

Solid start, looking forward to seeing what you do next. 🎉

commands/apps_test.go
nadivgold: namespace like the scoping in languages like c++ or would you just like me to change the name?

auxesis: Changing the name is perfect. 

That will make this test consistent with the rest of the code base.

commands/apps.go
```diff
@@ -293,49 +305,74 @@ func (c *AppsInitCmd) InitializeNodeBasicApp(stdout, stderr bytes.Buffer) (err e
...
 	defer checkPkgJSON.Close()
 	validPkgJSON, err := os.OpenFile("package.json", os.O_RDWR, 0777)
 	if err != nil {
-		panic(err)
+		return fmt.Errorf("failed to open package.json %w", err)
 	}
+	defer validPkgJSON.Close()
 	log.Println("[INFO] Validating package.json")
-	finfo, err := validPkgJSON.Stat()
+	buf, err := ioutil.ReadFile("package.json")
```
pavelnikolov: Rename with `os.ReadFile`. `ioutil` is deprecated in 1.16 which was released yesterday.

commands/apps.go
```diff
@@ -293,49 +305,74 @@ func (c *AppsInitCmd) InitializeNodeBasicApp(stdout, stderr bytes.Buffer) (err e
...
+			log.Println("[ERROR] unable to remove empty package.json")
+		}
+		log.Println("[WARN] package.json is empty. Creating package.json")
+		err = c.CreatePkgJson(stdout, stderr)
+		if err != nil {
+			log.Println("[ERROR] There was an error creating package.json. Is node installed?")
+			return fmt.Errorf("there was an error creating package.json. Is node installed? %w", err)
+		}
+		log.Println("[INFO] package.json created from empty file")
+		buf, err = ioutil.ReadFile("package.json")
```
pavelnikolov: `os.ReadFile` again

commands/apps.go
```diff
@@ -293,49 +305,74 @@ func (c *AppsInitCmd) InitializeNodeBasicApp(stdout, stderr bytes.Buffer) (err e
...
-	buf := make([]byte, finfo.Size())
-	_, err = validPkgJSON.Read(buf)
+	fStr := string(buf)
+	if len(fStr) == 0 {
+		err := os.Remove("package.json")
+		if err != nil {
+			log.Println("[ERROR] unable to remove empty package.json")
+		}
+		log.Println("[WARN] package.json is empty. Creating package.json")
+		err = c.CreatePkgJson(stdout, stderr)
```
pavelnikolov: (nitpick) `CreatePkgJSON` https://github.com/golang/go/wiki/CodeReviewComments#initialisms

commands/apps.go
```diff
@@ -293,49 +305,74 @@ func (c *AppsInitCmd) InitializeNodeBasicApp(stdout, stderr bytes.Buffer) (err e
...
+		if err != nil {
+			log.Println("[ERROR] There was an error creating package.json. Is node installed?")
```
pavelnikolov: We should not log and return an error - it's either one or the other. Wrap it and return it. Then some caller in the chain will log it.

commands/apps.go
```diff
@@ -293,49 +305,74 @@ func (c *AppsInitCmd) InitializeNodeBasicApp(stdout, stderr bytes.Buffer) (err e
...
+		if err != nil {
+			return fmt.Errorf("failed to read package.json %w", err)
+		}
+		fStr = string(buf)
+	}
+	jsonMap := make(map[string]interface{})
+	err = json.Unmarshal(buf, &jsonMap)
 	if err != nil {
-		panic(err)
+		log.Println("[ERROR] JSON format invalid for package.json")
```
pavelnikolov: Don't log and return an error. One or the other only

commands/apps.go
```diff
@@ -293,49 +305,74 @@ func (c *AppsInitCmd) InitializeNodeBasicApp(stdout, stderr bytes.Buffer) (err e
...
+		return fmt.Errorf("package.json is not valid JSON %w", err)
 	}
-	fStr := string(buf)
-	if !strings.Contains(fStr, `"start": "`) {
-		replace := `"scripts": {
-    "start": "node YOUR_SERVER_HERE.js",`
-		fStr = strings.Replace(fStr, `"scripts": {`, replace, 1)
-		err = validPkgJSON.Truncate(0)
+	lv := jsonMap["scripts"]
+	stringMap, ok := lv.(map[string]interface{})
```
pavelnikolov: Is there a better explanation than `stringMap`? Like files, lines, packages, etc?

commands/apps_test.go
```diff
@@ -45,3 +48,86 @@ func TestCommandsAppsCreateAttemptsToValidateStackOnError(t *testing.T) {
 	assert.Error(err)
 	assert.Regexp("bad request: unable to find stack", err)
 }
+
+// Looks like nested functions are not yet supported by the compiler
```
pavelnikolov: What do you mean by this comment?

commands/apps_test.go
```diff
@@ -45,3 +48,86 @@ func TestCommandsAppsCreateAttemptsToValidateStackOnError(t *testing.T) {
...
+func OverwriteFile(loc string, data string) (err error) {
+	err = os.Remove(loc)
```
pavelnikolov: Do you need to remove and recreate a file in order to rewrite it?

commands/apps.go
pavelnikolov: It will not error. It will return the default value, no? 
https://goplay.space/#Pewy3z5yqaT

commands/apps.go
```diff
@@ -293,49 +305,74 @@ func (c *AppsInitCmd) InitializeNodeBasicApp(stdout, stderr bytes.Buffer) (err e
 	checkPkgJSON, err := os.Open("package.json")
 	if err != nil {
 		log.Println("[WARN] package.json does not exist. Creating package.json")
-		cmd := exec.Command("npm", "init", "-y")
-		cmd.Stdout = &stdout
-		cmd.Stderr = &stderr
-		err := cmd.Run()
+		err := c.CreatePkgJson(stdout, stderr)
 		if err != nil {
 			log.Println("[ERROR] There was an error creating package.json. Is node installed?")
```
pavelnikolov: Remove the log if you return here

commands/apps_test.go
nadivgold: Huh, looks like when I looked up nested function in go the top result was [this](https://stackoverflow.com/questions/21961615/what-are-the-problems-that-are-mitigated-by-not-allowing-nested-function-declara) stack overflow post from 2017. 

Since then it does look like nested functions are now [possible](https://sharbeargle.gitbooks.io/golang-notes/content/nested-functions.html). The comment was an explanation of why that function was in the main scope rather than the scope of the test.

commands/apps.go
```diff
@@ -241,14 +243,24 @@ func (c *AppsInitCmd) Run() (err error) {
 	case "nodejs-basic":
 		err := c.InitializeNodeBasicApp(stdout, stderr)
 		if err != nil {
-			panic(err)
+			log.Printf("[ERROR]: init completed with error %x", err)
```
pavelnikolov: Don't log and return the error

commands/apps_test.go
nadivgold: good note, no I don't.

auxesis (approved):
I'm happy with the changes you've made @nadivgold.

PR #104, opened 2021-02-26
Make tests run on Windows

auxesis:
Also test Windows-specific file naming behavior.

### DIFF
--- a/sectionctl_test.go
+++ b/sectionctl_test.go
@@ -1,0 +1,32 @@
+package main
+
+import (
+	"io/ioutil"
+	"net/url"
+	"path/filepath"
+	"testing"
+
+	"github.com/alecthomas/kong"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestBootstrapSetsUpDebugLogFile(t *testing.T) {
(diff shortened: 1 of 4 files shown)


sectionctl_test.go
```diff
@@ -0,0 +1,32 @@
...
+	assert := assert.New(t)
+
+	// Setup
+	u, err := url.Parse("https://aperture.section.io")
+	assert.NoError(err)
+	d, err := ioutil.TempDir("", "sectionctl")
+	assert.NoError(err)
+
+	var cli = CLI{Debug: true, SectionToken: "s3cr3t", SectionAPIPrefix: u, DebugFileDir: d}
+	var ctx = kong.Context{}
```
pavelnikolov: (nitpick) same as the default value:

```go
var ctx kong.Context
```

pavelnikolov (approved):

AskAlice (approved):
thank you!

sectionctl_test.go
auxesis: Good call. It was being initialized with a value, but I refactored it so it didn't.

PR #106, opened 2021-03-11
error messages about adding sectionctl_token to path now work for wsl2.

nadivgold:
Error messages now work for wsl2

co-authored with @AskAlice

### DIFF
--- a/commands/login.go
+++ b/commands/login.go
@@ -19,23 +19,28 @@ type LoginCmd struct {
 
 // Run executes the command
 func (c *LoginCmd) Run() (err error) {
+	screenshot := "https://raw.githubusercontent.com/section/sectionctl/main/docs/section_token_control_panel.png"
+	windowsStr := fmt.Sprintf("Unable to write credential.\n\nPlease execute the following, add it to your Powershell profile, or add it to your environment variables in control panel: \nWith Powershell:\n$env:SECTION_TOKEN=\"%s\"\n\nWith CMD:\nset SECTION_TOKEN=%s\n\nWith control panel:\n%s", api.Token, api.Token, screenshot)
+	linuxStr := fmt.Sprintf("Unable to write credential.\n\nPlease run this command, and add it to your ~/.bashrc (you do not need to run sectionctl login again)\n\nexport SECTION_TOKEN=%s", api.Token)
 	if api.Token != "" {
 		err = credentials.Write(api.PrefixURI.Host, api.Token)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


AskAlice:
I would rather have @pavelnikolov or @mwickman review this since I half-wrote it.

we tested it and it works, though.

commands/login.go
```diff
@@ -19,23 +19,28 @@ type LoginCmd struct {
 
 // Run executes the command
 func (c *LoginCmd) Run() (err error) {
+	screenshot := "https://raw.githubusercontent.com/section/sectionctl/main/docs/section_token_control_panel.png"
+	windowsStr := fmt.Sprintf("Unable to write credential.\n\nPlease execute the following, add it to your Powershell profile, or add it to your environment variables in control panel: \nWith Powershell:\n$env:SECTION_TOKEN=\"%s\"\n\nWith CMD:\nset SECTION_TOKEN=%s\n\nWith control panel:\n%s", api.Token, api.Token, screenshot)
+	linuxStr := fmt.Sprintf("Unable to write credential.\n\nPlease run this command, and add it to your ~/.bashrc (you do not need to run sectionctl login again)\n\nexport SECTION_TOKEN=%s", api.Token)
 	if api.Token != "" {
 		err = credentials.Write(api.PrefixURI.Host, api.Token)
 		if err != nil {
```
pavelnikolov, line 27: I'd rather use a switch statement based on `runtime.GOOS` and have separate cases for `"linux", "darwin"` and `windows`.

commands/login.go
```diff
@@ -19,23 +19,28 @@ type LoginCmd struct {
...
-			if runtime.GOOS == "linux" || runtime.GOOS == "darwin" {
-				fmt.Printf("Unable to write credential.\n\nPlease run this command, and add it to your ~/.bashrc\n\nexport SECTION_TOKEN=%s\n",api.Token)
-				return nil
```
pavelnikolov, line 27: Aren't you loosing this case completely?

commands/login.go
AskAlice: Why? Linux and Mac return the same string.

commands/login.go
AskAlice: yes it returns lower down

commands/login.go
pavelnikolov: Yes, but it'd be more obvious - just like in the if/else statement before. Now we only check for windows and I had to read it twice to understand it.

pavelnikolov (approved):

PR #107, opened 2021-03-15
Deploy using git instead of Patch so that we get the git prereceive v…

mwickman:
…alidation back and have a longer timeout on the operation

Credit to @AskAlice for this git code, I just cleaned up some errors/testing mostly.


commands/gitService.go
```diff
@@ -0,0 +1,144 @@
...
+// GitService interface provides a way to interact with Git
+type GitService interface {
+	UpdateGitViaGit(c *DeployCmd, response UploadResponse) error
+}
+
+// GS ...
+type GS struct{}
+
+// This is far less then ideal, however Kong does not seem to provide a way to inject dependencies into its commands so we must use this for testing
+var globalGitService GitService = &GS{}
```
mwickman, line 25: If someone knows another way to do DI with these Kong commands I'd be happy to change this.

commands/ps.go
```diff
@@ -1,115 +1,115 @@
-package commands
```
AskAlice, line 1: CRLF issue?

commands/ps.go
```diff
@@ -1,115 +1,115 @@
...
-				a.PayloadID,
-			}
-			table.Append(r)
-		}
-	}
-
-	table.Render()
-	return nil
-}
+package commands
```
mwickman, line 1: This file only has line endings changed after formatting

commands/logs.go
```diff
@@ -1,90 +1,90 @@
...
-			startTimestampRfc3339 = t.Format(time.RFC3339)
-		}
-		s.Prefix = ""
-		s.FinalMSG = ""
-		s.Start()
-		time.Sleep(5 * time.Second)
-	}
-	return nil
-}
+package commands
```
mwickman, line 1: This file only has line endings changed after formatting

AskAlice:
potentially, could we make it a flag to use the old deploy if customers run into issues?

commands/deploy.go
```diff
@@ -31,7 +31,7 @@ type DeployCmd struct {
 	AppID          int           `required short:"i" help:"AppID to deploy application to."`
 	Environment    string        `short:"e" default:"production" help:"Environment to deploy application to."`
 	Debug          bool          `help:"Display extra debugging information about what is happening inside sectionctl."`
-	Directory      string        `short:"C" default:"." help:"Directory which contains the application to deploy."`
```
AskAlice: why?

AskAlice (commented):
testing this locally, but looks good from what I can tell.

commands/deploy.go
mwickman: I think this was copy-pasted from your PR. Seems like we could leave the shorthand flag in place, I'll add it back

mwickman:
> potentially, could we make it a flag to use the old deploy if customers run into issues?

If they run into issues they could downgrade sectionctl. I'm hesitant to add the option because ideally, we don't want to support both. I've been able to deploy well from Linux, and assuming you have been testing via Windows I'm feeling like these are good changes.

AskAlice:
interesting output from the git hook pre-receive. should hide that output in one of our private repos.

also we should probably remove the part where it says `running npm ci and npm run build` in our nodejs module because it doesn't actually do that in this context.

image

AskAlice (approved):

commands/deploy.go
AskAlice: ah. yeah it probably was giving me some sort of error. not sure if it was at a local dev, compilation, or runtime level issue though.

AskAlice:
theoretically we should also make sure to remove the repo from the temp folder when we're done with it.

commands/deploy.go
```diff
@@ -150,18 +150,8 @@ func (c *DeployCmd) Run() (err error) {
...
-			Op: "replace",
-			Value: PayloadValue{
-				ID: response.PayloadID,
-			},
-		},
-	}
-	err = api.ApplicationEnvironmentModuleUpdate(c.AccountID, c.AppID, c.Environment, c.AppPath+"/.section-external-source.json", ups)
-	s.Stop()
+
+	err = globalGitService.UpdateGitViaGit(c, response)
```
pavelnikolov, line 154: Don't you need to use context?

commands/logs.go
pavelnikolov: We need a linter for that

PR #108, opened 2021-03-16
add author to git commit

AskAlice:
if a user doesn't have a git author set globally, it would error. This should fix that.

Tested and it commits with the explicitly specified author now.

### DIFF
--- a/commands/gitService.go
+++ b/commands/gitService.go
@@ -7,10 +7,11 @@ import (
 	"log"
 	"os"
 	"path/filepath"
-
+	"time"
 	"github.com/go-git/go-git/v5"
 	gitHTTP "github.com/go-git/go-git/v5/plumbing/transport/http"
 	"github.com/section/sectionctl/api"
+	"github.com/go-git/go-git/v5/plumbing/object"
 )
 
 // GitService interface provides a way to interact with Git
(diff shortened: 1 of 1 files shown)


mwickman (approved):

commands/gitService.go
```diff
@@ -112,7 +113,11 @@ func (g *GS) UpdateGitViaGit(c *DeployCmd, response UploadResponse) error {
 	if err != nil {
 		return err
 	}
-	commitHash, err := w.Commit("[sectionctl] updated nodejs/.section-external-source.json with new deployment.", &git.CommitOptions{})
+	commitHash, err := w.Commit("[sectionctl] updated nodejs/.section-external-source.json with new deployment.", &git.CommitOptions{Author: &object.Signature{
+		Name:  "sectionctl",
+		Email: "noreply@section.io",
```
jstangroome, line 118: Can we obtain their actual name and email address from their API token using an extra API call?
e.g. https://aperture.section.io/api/ui/#!/User/currentUserGet

AskAlice: oh we definitely could. The previous implementation the commit defaulted to global git settings, and the implementation before that (a PATCH request to our backend) also hardcoded a fake machine user.

PR #115, opened 2021-04-20
Remove requirement for server.conf to be present for the Node.js Module

jstangroome:
The Node.js Module itself removed the requirement for this file in December 2020.

We should also remove the creation of server.conf during `apps init` in a future PR.


AskAlice (approved):
LGTM

PR #116, opened 2021-04-20
Do not create server.conf if missing but validate if present

jstangroome:
Since December 2020 the Node.js Module does not require the server.conf file and it is considered deprecated.

Continues from PR #115


AskAlice (approved):
LGTM

PR #117, opened 2021-05-12
Feature/ci mode

nadivgold:
Add a CI mode for `deploy` that suppresses non-error console messages. Uses `--quiet` flag or the "CI" environment variable.


AskAlice (approved):

commands/deploy.go
```diff
@@ -70,9 +70,18 @@ func (c *DeployCmd) Run() (err error) {
 			return fmt.Errorf("not a valid Node.js app: \n\n%s", errstr)
 		}
 	}
+	
+	
+	envCI, err := strconv.ParseBool(os.Getenv("CI"))
```
mwickman, line 75: I think we should make use of Kong's env variable features rather than writing our own code to handle it, as the example here does https://github.com/section/sectionctl/blob/8c894a8f722c81ee2f2232aee5ee8be756cb0e3f/sectionctl.go#L37. This will help us keep naming the same. I'm actually not even sure if we should use a environment variable for this 'quiet' feature, as in a CI tool someone can just set the flag on the commands. `CI` is a very generic environment variable name, what if someone is using that in their own system and we unintentionally pick it up? What do you think about just going with the flag and no env variable?

commands/gitService.go
```diff
@@ -40,20 +40,33 @@ func (g *GS) UpdateGitViaGit(c *DeployCmd, response UploadResponse) error {
...
 	payload := PayloadValue{ID: response.PayloadID}
 	branchRef := fmt.Sprintf("refs/heads/%s",c.Environment)
-	r, err := git.PlainClone(tempDir, false, &git.CloneOptions{
-		URL:      fmt.Sprintf("https://aperture.section.io/account/%d/application/%d/%s.git", c.AccountID, c.AppID, appName),
-		Auth:     gitAuth,
-		Progress: os.Stdout,
-		ReferenceName: plumbing.ReferenceName(branchRef),
-	})
+	var r *git.Repository
+	if err != nil {
```
mwickman, line 52: This doesn't appear to be checking any error

commands/gitService.go
```diff
@@ -40,20 +40,33 @@ func (g *GS) UpdateGitViaGit(c *DeployCmd, response UploadResponse) error {
...
+		return err
+	}
+	if c.Quiet {
```
mwickman, line 55: The only thing we really need to change is what goes in the Progress field it looks like. Putting the entire thing in an if/else duplicates much more code than necessary. Something like:
```
p := os.Stderr
if c.Quiet { p = nil }
```
And then passing `p` to the git.PlainClone call would save many lines.

sectionctl.go
```diff
@@ -46,9 +46,10 @@ func bootstrap(c CLI, ctx *kong.Context) {
 
 	colorableWriter := colorable.NewColorableStderr()
 
+	minLogLevel := logutils.LogLevel("INFO")
```
mwickman, line 49: This change doesn't appear to be doing anything to me. The Quiet flag should change our minLogLevel so that we don't have to check the status of c.Quiet when we are logging things.

commands/deploy.go
```diff
@@ -156,7 +167,9 @@ func (c *DeployCmd) Run() (err error) {
 		return fmt.Errorf("failed to trigger app update: %v", err)
 	}
 
-	fmt.Println("Done!")
+	if !c.Quiet{
+		fmt.Println("Done!")
```
mwickman, line 171: We should use the `log` interface here so that we don't have to check c.Quiet, it will just be handled automatically by the set log level.

commands/deploy.go
```diff
@@ -131,7 +140,9 @@ func (c *DeployCmd) Run() (err error) {
 	artifactSizeMB := stat.Size() / 1024 / 1024
 	log.Printf("[DEBUG] Upload artifact is %dMB (%d bytes) large", artifactSizeMB, stat.Size())
 	s = NewSpinner(fmt.Sprintf("Uploading app (%dMB)...", artifactSizeMB))
-	s.Start()
+	if !(c.Quiet){
```
mwickman, line 143: Instead of checking Quiet when we start the spinner, I think a better pattern would be for us to modify our `NewSpinner` function to return a no-op spinner that matches the `spinner.Spinner` interface, but does nothing when start/stop are called. Then we don't have to remember to make this check if we add more spinners in the future.

commands/deploy.go
nadivgold: What if we set it to SECTION_CI env var?

PR #119, opened 2021-05-18
Context object added to track context within sectionctl. 

nadivgold:
added `context.Context` to each file and function call, so that context can be kept. Context now tracks quiet mode and enables minimum output when the flag is raised

I've kept this on the feature/CI branch since the only tangible addition is the CI mode

### DIFF
--- a/commands/apps_test.go
+++ b/commands/apps_test.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"context"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -39,8 +40,11 @@ func TestCommandsAppsCreateAttemptsToValidateStackOnError(t *testing.T) {
 		StackName: "helloworld-1.0.0",
 	}
 
+	// Define Contet
+	ctx := context.Background()
+
 	// Invoke
-	err = cmd.Run()
+	err = cmd.Run(ctx)
 
 	// Test
 	assert.True(stackCalled)
@@ -107,8 +111,13 @@ func TestCommandsAppsInitHandlesErrors(t *testing.T) {
 				fmt.Println("server.conf creation failed")
 			}
 
+
+			// Define Context
+
+			ctx := context.Background()
+
 			// Invoke
-			err = cmd.Run()
+			err = cmd.Run(ctx)
 
 			// Test
 			if tc.isFatal {
--- a/commands/apps_test.go
+++ b/commands/apps_test.go
@@ -40,7 +40,6 @@ func TestCommandsAppsCreateAttemptsToValidateStackOnError(t *testing.T) {
 		StackName: "helloworld-1.0.0",
 	}
 
-	// Define Contet
 	ctx := context.Background()
 
 	// Invoke
@@ -110,9 +109,6 @@ func TestCommandsAppsInitHandlesErrors(t *testing.T) {
 			if err != nil {
 				fmt.Println("server.conf creation failed")
 			}
-
-
-			// Define Context
 
 			ctx := context.Background()
 
--- a/commands/helpers.go
+++ b/commands/helpers.go
@@ -1,7 +1,9 @@
 package commands
 
 import (
+	"context"
 	"fmt"
+	"io"
 	"os"
 	"time"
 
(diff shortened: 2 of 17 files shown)


commands/apps_test.go
```diff
@@ -39,8 +40,11 @@ func TestCommandsAppsCreateAttemptsToValidateStackOnError(t *testing.T) {
 		StackName: "helloworld-1.0.0",
 	}
 
+	// Define Contet
```
pavelnikolov: (nitpick) this comment has a typo. In fact it is not needed at all.

commands/helpers.go
```diff
@@ -15,10 +17,20 @@ func NewSpinner() *spinner.Spinner {
 }
 */
 
+type CTXKEY string
```
pavelnikolov: Avoid using all caps in Golang

sectionctl.go
```diff
@@ -38,15 +39,25 @@ type CLI struct {
...
 
-func bootstrap(c CLI, ctx *kong.Context) {
+
+
+
+func bootstrap(c CLI, ctx *kong.Context) context.Context {
 	api.PrefixURI = c.SectionAPIPrefix
 	api.Timeout = c.SectionAPITimeout
 
+	contxt := context.Background()
```
pavelnikolov: use `ctx` instead. Don't use the name of a package as variable. The auto-complete doesn't work after this line for the `context` package.

sectionctl.go
```diff
@@ -38,15 +39,25 @@ type CLI struct {
...
+	contxt = context.WithValue(contxt, commands.CTXKEY("quiet"), c.Quiet)
+
 	colorableWriter := colorable.NewColorableStderr()
 
-	minLogLevel := logutils.LogLevel("INFO") 
+	minLogLevel := logutils.LogLevel("INFO")
+	if contxt.Value(commands.CTXKEY("quiet")).(bool) {
+		minLogLevel = logutils.LogLevel("ERROR")
+	}
```
pavelnikolov, line 58: You can rewrite the first line of this func like this:

```go
func bootstrap(c CLI, ctx *kong.Context) context.Context {
	api.PrefixURI = c.SectionAPIPrefix
	api.Timeout = c.SectionAPITimeout

	ctx := context.Background()
	minLogLevel := logutils.LogLevel("INFO")
	if c.Quiet {
		ctx = context.WithValue(contxt, commands.CTXKEY("quiet"), c.Quiet)
		minLogLevel = logutils.LogLevel("ERROR")
	}

	colorableWriter := colorable.NewColorableStderr()
```

sectionctl.go
```diff
@@ -102,8 +114,9 @@ func main() {
 		kong.UsageOnError(),
 		kong.ConfigureHelp(kong.HelpOptions{Tree: true}),
 	)
-	bootstrap(cli, ctx)
-	err := ctx.Run()
+	contxt := bootstrap(cli, ctx)
```
pavelnikolov: Don't use `context` as variable name.

sectionctl.go
```diff
@@ -102,8 +114,9 @@ func main() {
...
+	ctx.BindTo(contxt, (*context.Context)(nil))
```
pavelnikolov: Why would you ever want to pass a nil context? I don't understand why this is needed.

sectionctl.go
pavelnikolov: Call this `ctx` and the other one `kongCtx` to make a distinction.

sectionctl.go
nadivgold: `ctx` is used on line 112 for the Kong parsing context. I was hesitant to replace that, but I can rename it so I can use `ctx` as context.Context in every file

sectionctl.go
nadivgold: I've renamed the kong `ctx` to `cmd`, since that is more accurate to the information that it holds.

sectionctl.go
nadivgold: I'm not sure exactly what `BindTo` does, but it does not work without it. This was how code on the internet. My best guess it that it creates a reference to the context object so that Kong passes the context by reference. This would make mutations affect the base object.

AskAlice (approved):

pavelnikolov:
Maybe we should have used a boolean value that we pass around instead of passing a value through the context, to be honest. We don't expect to add many more values like this and the context was not created for that. Not a show-stopper but I think that this sacrifices simplicity.

PR #127, opened 2021-06-04
feat: basic-auth

luft-ecommerce:


AskAlice:
we could implement this into sectionctl natively as long as it's optional and not enabled by default. something like
```sectionctl login --use-basic-auth```

Then it could store your credentials using github.com/zalando/go-keyring (in https://github.com/section/sectionctl/blob/main/credentials/credentials.go)

luft-ecommerce:
It would be great if we can use Basic Authentication to generate certificate (`sectionctl certs`) and deploy (`sectionctl deploy`).

AskAlice:
@luft-ecommerce Could you clarify the use-case for this? Why prefer basic auth over tokens?

luft-ecommerce:
Hi @AskAlice, for us token would be the preferable method, but unfortunately, it can only grant access to the apps and envs, that were created before the token itself was generated. 
This is a blocker for our use case because we try to create, manipulate (and hopefully soon delete) apps and envs dynamically.

Generally, this is the main reason for switching to basic.

AskAlice:
hi @luft-ecommerce, we have updated our backend to update the permissions for apikeys when a user is added/removed from an account. We have also updated all user permissions on all apikeys to get them all in sync.

This should make this implementation and use-case redundant and you should be good to be running the upstream version of sectionctl.

PR #129, opened 2021-06-07
add cross-platform build pipeline for deb, rpm, homebrew, and windows installer

AskAlice:
with git actions it will package everything up, and add to the release

Windows installer with [NSIS](https://nsis.sourceforge.io/Main_Page)
deb package, rpm package, homebrew tap 

mac install (also works with linuxbrew)
```
brew tap section/brews
brew install sectionctl
```

will add to winget, a ppa, and a yum repo later

see the artifacts of this git action to see the results:
https://github.com/section/sectionctl/actions/runs/916207729#artifacts


commands/parsePackageJson.go
```diff
@@ -84,3 +86,42 @@ func ParseSectionConfig(sectionConfigContents string) (SectionConfigJSON, error)
...
 }
+
+// JSON returns a Resolver that retrieves values from a JSON source.
+//
+// Hyphens in flag names are replaced with underscores.
+func PackageJSONResolver(r io.Reader) (kong.Resolver, error) {
+	buf := new(bytes.Buffer)
+	_, err := buf.ReadFrom(r)
+	if err != nil{
+		return nil,nil
```
mwickman, line 97: We should return errors so they can be handled, or at least handle it by logging if that's what we want to happen.

commands/parsePackageJson.go
```diff
@@ -84,3 +86,42 @@ func ParseSectionConfig(sectionConfigContents string) (SectionConfigJSON, error)
...
+				return packageJSON.Section.AppID, nil
+			}
+		}
+		if len(packageJSON.Section.Environment) > 0 && flag.Name=="environment" {
+			return packageJSON.Section.Environment, nil
+		}
+		if len(packageJSON.Section.ModuleName) > 0 && flag.Name=="app-path" {
+			return packageJSON.Section.ModuleName, nil
+		}
+		return nil, nil
```
mwickman, line 123: If we don't want this function to return errors then lets remove that from the return values above, otherwise we should return errors when we have them.

.github/workflows/release.yml
```diff
@@ -25,57 +25,242 @@ jobs:
...
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      with:
-        upload_url: ${{ needs.release.outputs.release_upload_url }}
-        asset_path: dist/sectionctl-${{ env.VERSION }}-${{ env.GOOS }}-${{ env.GOARCH }}.tar.gz
-        asset_name: sectionctl-${{ env.VERSION }}-${{ env.GOOS }}-${{ env.GOARCH }}.tar.gz
-        asset_content_type: application/gzip
+            Created by https://github.com/mislav/bump-homebrew-formula-action
+        env:
+          COMMITTER_TOKEN: ${{ secrets.COMMITTER_TOKEN }}
+          # GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
mwickman, line 266: Remove commented line?

sectionctl.go
```diff
@@ -19,6 +19,7 @@ import (
 	"github.com/willabides/kongplete"
 )
 
+//go:generate go-winres make --product-version=git-tag --file-version=git-tag
```
mwickman, line 22: What is this line doing?

mwickman (approved):
Most of the build stuff is too complex and unfamiliar for me to really review it. Seems ok as long as you test each of the pipelines after merging and fix the go error handling. Let's be careful about putting too much more time towards this as we have other goals to be accomplishing right now.

commands/parsePackageJson.go
AskAlice: this is kong internals
it's basically https://github.com/alecthomas/kong/blob/master/resolver.go#L33

If it can't read the package json, then it will get the flags from the user input

commands/parsePackageJson.go
AskAlice: https://github.com/alecthomas/kong/blob/master/resolver.go#L33

sectionctl.go
AskAlice: this is windows-specific binary metadata that you can generate via `go generate`

AskAlice (commented):
> Most of the build stuff is too complex and unfamiliar for me to really review it. Seems ok as long as you test each of the pipelines after merging and fix the go error handling. Let's be careful about putting too much more time towards this as we have other goals to be accomplishing right now.

Yep I have tested the debs, the RPMs (on a centos droplet), homebrew, and installer

PR #136, opened 2021-07-27
Work by default

lcaballero:
Adding checks for required executables.  The makefile now gives warnings if the executables are not found.


Makefile
```diff
@@ -6,14 +6,29 @@ export HEAPAPPID := 892134159
...
+GO_VERSION := $(shell go version | cut -d' ' -f3)
 
 all: test
 
+check_environment:
+ifneq ("${GO_VERSION}", "go1.16.6")
+	$(error "requires go version 1.16.6 or higher")
+endif
+ifndef HAS_STATICCHECK
+	$(error "staticcheck is not on PATH")
```
jstangroome, line 25: Perhaps we should declare required tools in `go.mod`, e.g.:
https://marcofranssen.nl/manage-go-tools-via-go-modules#tldr

p2rbgvh6p6-lang: Okay I was almost sure as figured it out

PR #138, opened 2021-08-13
Remove Apps Init from Sectionctl

nadivgold:
`apps init` is being deprecated and replaced by [sectionctl-helper](https://github.com/section/sectionctl-helper).


nadivgold:
this change is not urgent and does not need its own version bump; it can wait until the next cut of sectionctl.

AskAlice (approved):

PR #139, opened 2021-08-17
improved prevalidation

nadivgold:
Added some more prevalidation protections to stop broken node apps from being uploaded.

### DIFF
--- a/commands/deploy.go
+++ b/commands/deploy.go
@@ -178,7 +178,10 @@ func IsValidNodeApp(dir string) (errs []error) {
 		if err != nil {
 			log.Info().Err(err).Msg("Error reading your package.json")
 		}
-		packageJSON, _ := ParsePackageJSON(string(packageJSONContents))
+		packageJSON, err := ParsePackageJSON(string(packageJSONContents))
+		if err != nil {
+			log.Info().Err(err).Msg("Error parsing your package.json")
+		}
 		if len(packageJSON.Section.StartScript) == 0 && packageJSON.Scripts["start"] == "" {
 			errs = append(errs, fmt.Errorf("package.json does not include a start script. please add one"))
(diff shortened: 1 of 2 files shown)


AskAlice (approved):
tested locally, works as intended

commands/deploy.go
```diff
@@ -171,9 +171,24 @@ func (c *DeployCmd) Run(ctx *kong.Context, logWriters *LogWriters) (err error) {
...
 	packageJSONPath := filepath.Join(dir, "package.json")
-	if _, err := os.Stat(packageJSONPath); os.IsNotExist(err) {
-		errs = append(errs, fmt.Errorf("%s is not a file", packageJSONPath))
+	if _, err := os.Open(packageJSONPath); os.IsNotExist(err) {
+		log.Debug().Msg(fmt.Sprintf("[WARN] %s is not a file", packageJSONPath))
+	} else {
+		packageJSONContents, err := ioutil.ReadFile(packageJSONPath)
+		if err != nil {
+			log.Info().Err(err).Msg("Error reading your package.json")
+		}
```
jstangroome: Calling `os.Open` and discarding the returned file handle is a resource leak. Just check if any error from `ioutil.ReadFile` represents a non-existent file. Try something like this
```suggestion
	if packageJSONContents, err := ioutil.ReadFile(packageJSONPath); err != nil {
		if os.IsNotExist(err) {
			log.Debug().Msg(fmt.Sprintf("[WARN] %s is not a file", packageJSONPath))
		} else {
			log.Info().Err(err).Msg("Error reading your package.json")
		}		
	} else {
```

AskAlice: yeah i felt like there might have been a better solution to that. thanks

mwickman (approved):

